PR #9, opened 2025-09-19
TE-4

SkyBeal:
Player possession in. Camera switches and player not allowed to move while in vase. Added player, vase, and input manager prefabs. All found in Player Possession scene.


bboster (changes requested):
Take a look again and see why player cannot move in Player Possession scene. 
If extra steps are needed to test, please put them in the description.

bboster (approved):
Looks great!

JBBateman (approved):
Everything seems to work fine on my end.

PR #10, opened 2025-09-19
TE-5

JBBateman:
- Created Guard Prefab
- Created PatrolPath Prefab
- Guard patrols a path
- Guard leaves path to chase player when player enters vision cone
- Guard abandons chase and return to patrol if player possesses an object
- Guard cannot detect the player when they are possessing an object
- Guard prints "Player Caught" in the console when successfully catching the player, then returns to its patrol path
- Guard abandons chase if player leaves vision cone


bboster (approved):
In the future let's rename every instance of enemy to guard.
Looks good other than that, approved.

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/Behavior.cs
```diff
@@ -0,0 +1,57 @@
...
+        currentLoop = null;
+    }
+
+    #endregion
+
+    /// <summary>
+    /// Controls the overall flow of the behavior.
+    /// </summary>
+    /// <returns></returns>
+    protected virtual IEnumerator BehaviorLoop()
```
TheBeesSneeze, line 48: could these be abstract actually?

Geist Heist/Assets/Scripts/Player/Movement/FirstPersonInputHandler.cs
```diff
@@ -63,6 +66,7 @@ public override void OnPossessStarted()
             if (results.transform.GetComponent<PossessableObject>() && results.transform != this.transform)
             {
                 PlayerManager.Instance.PossessObject(results.transform.GetComponent<PossessableObject>());
+                OnPossessObject?.Invoke(0);
```
TheBeesSneeze, line 69: what does 0 mean here

TheBeesSneeze (approved):
very small stuff, looks good

PR #26, opened 2025-09-25
Te 13

JRavioli:
Added scripts for sound systems and audio manager.
Sound systems are ready to start testing.
Did NOT make audio prefabs yet as I want to test these systems before putting them anywhere else, I will need programmers help with this so I can test 3D and looping sfx with the 1st person perspective for now.  I also want to know what exactly the 3rd person perspective looks like so I can play around with attenuation.


bboster (approved):
I trust you

TheBeesSneeze (approved):
thanks for teaching me fmod i actually know whats going on here

PR #29, opened 2025-09-30
OOO-12

bboster:
Added Globe, Level 1 and Lobby scenes. Created a transition between the three of them in FirstPersonInputHandler.cs


TheBeesSneeze (changes requested):
If it works it works, (but there do be some style changes, alas)

Geist Heist/Assets/Scripts/Player/Movement/FirstPersonInputHandler.cs
```diff
@@ -32,9 +40,9 @@ void Start()
     }
 
     // Update is called once per frame
-    void Update()
+    void FixedUpdate()
     {
-
+        TransitionScene();
```
TheBeesSneeze, line 45: can this be moved to Update() or LateUpdate() ?
FixedUpdate is going to call this function multiple times per frame which is fine for now, but could build up to add some extra lag later. 

tbh i have this really good modular interaction script that i stole from marissa moser that does this stuff really well. if you want, i could implement that tomorrow in class (and u can just not worry about this for now)

TheBeesSneeze (changes requested):
forgot one thing

Geist Heist/Assets/Scripts/Player/Movement/FirstPersonInputHandler.cs
```diff
@@ -23,6 +24,13 @@ public class FirstPersonInputHandler : IInputHandler
     [SerializeField] private float sphereCastRadius = 10;
     [SerializeField] private float sphereCastDistance = 100000000;
 
+    // Scene transition specific variables
+    [SerializeField] private string sceneName;
```
TheBeesSneeze, line 28: okay, thinking about it more, this method of switching scenes may not work if we are doing a hubworld (where there can be multiple scenes that can be transitioned to)
instead of storing the scene to transition to on the player, maybe we should store it on the door objects (and make a door script)

TheBeesSneeze (changes requested):
ok wait sorry i thought of something else

Geist Heist/Assets/Scripts/Player/Movement/FirstPersonInputHandler.cs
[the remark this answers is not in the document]
bboster: kk on it

bboster: this is just for the first milestone. it works for this version. we can refractor later.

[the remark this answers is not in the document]
bboster: done

TheBeesSneeze (approved):
Hell yeah, looks good, ty Brookie!!!

jtkelly18 (approved):

PR #32, opened 2025-10-02
Ooo 7

BBurtz:


SkyBeal (approved):
we might want to orient the can differently so that it rolls, but overall everything functions

important to note that the guard behavior isn't implemented with the can yet but i'm sure jacob is working on that

Geist Heist/Assets/Scripts/Player/Movement/Input Maps/Player Input.inputactions
bboster, line 1: Why was this set to q and not e?

bboster (approved):
APPROVED -knuckles

PR #33, opened 2025-10-03
Ooo 10

TheBeesSneeze:
Optional collectables, tether possessables, and JSON file saving / persisting save data.

To test this, id recommend playing the lobby scene, then looking at 'collectable testing' (collectable testing will bring you back to the lobby)


bboster (changes requested):
please put how to change render pipeline in the description of this. current shaders are non pink. many people may need to change things on their machine :))

bboster (commented):
Please move the tether to level one

bboster (approved):
approved

SkyBeal (approved):
everything listed is working!

only current problem i see is going back through the brown cube in level 1 drops the gate even if you don't get the tether BUT the level shouldn't be left without it anyway. just something to keep in mind for the future.

PR #34, opened 2025-10-03
Ooo 27

TheBeesSneeze:
Fixed bug with player input direction stuff. Also the camera wont clip into stuff as much now.


bboster (approved):
APPROVED -knuckles

SkyBeal (approved):
we should bump the speed up but consider the movement FIXED!

PR #35, opened 2025-10-04
OOO-8

Veizyl:
Finished the cooldown bar.
To test you can go to CooldownBar scene found in Scenes -> Testing
To implement into other scenes you would drag the CooldownManager prefab (Prefabs -> Constants) and the CooldownCanvas (Prefabs -> Player -> Possession) into the scene. 
Drag the CooldownBar from the CooldownCanvas into the CooldownSlider field in CooldownManager.
Drag the Cooldown Canvas and CooldownManager into the fields on the 3rd person player(Third Person Input Handler)
Should default to 5 seconds.
Does not actually prevent the player from repossessing the object while the cooldown is running.
For new objects made in the OnInteractStarted() function there would need to be a line added that says             "CooldownManager.Instance.StartCooldown();" like in the VaseObject script.


bboster (changes requested):
Right now the player can still possess the vase even when the cooldown bar is showing. Please dig into this.

bboster (approved):
APPROVED! -knuckles

TheBeesSneeze (changes requested):
Looks really good, just asking u to incorporate not possessing stuff while the cooldown is active, cuz i think it shouldnt be too complicated.

TheBeesSneeze (changes requested):
For the cooldown canvas, please make sure that its scale mode is set to scale with screen size, 1920x1080

TheBeesSneeze (approved):
Yay! ty looks good

PR #36, opened 2025-10-04
OOO-14

JBBateman:
Guard behaviors are done. It should react to everything except the possessable objects and their associated effects. Another note is that the vision cone will still detect the player through walls, I haven't found a good solution to replace the current iteration with yet.


bboster (approved):
APPROVED -knuckles

jtkelly18 (approved):
Amazing Work

PR #37, opened 2025-10-06
OOO - 6

SkyBeal:
-made toy car prefab
-tap for small movement, hold for longer movement (with a cap)
-cannot move car without possession
-ghost unpossesses near the car
-cannot spam movement

best to test in player possession third person scene


jtkelly18 (approved):
approved :)

bboster (approved):
APPROVED

PR #39, opened 2025-10-07
OOO-45

JBBateman:
Guard now reloads the scene when the player gets caught.


bboster (approved):
APPROVED -knuckles

jtkelly18 (approved):
works well

PR #41, opened 2025-10-07
OOO_49

Navahkiin:
Adding temp possessable models + textures as well as the other temp models and textures that were in the drive


SkyBeal (approved):
design will probably have to make some new temp textures but looks good!

bboster (approved):
approved

PR #43, opened 2025-10-07
OOO-44

jtkelly18:
Updated Vending Machine prefab to add the reference to the cinemachine camera child of the vending machine object


SkyBeal (approved):
works 👍 I think there's another MR open that drags all the constants into the scene so this one's fine without them

bboster (approved):
approved

PR #49, opened 2025-10-08
Ooo 41

Afederspiel:
Vending machine shoots, Guards still don't move on my end yet, I'm not sure what bug is causing it or if I needed to add something.


smnoonan (approved):

bboster (approved):

PR #56, opened 2025-10-09
OOO 67

BBurtz:


TheBeesSneeze (approved):
This may have to be redone a little once the billboard UI system gets in but yeah it works for now

JBBateman (approved):
The vending machine doesn't charge but I think I remember that being covered in class. Plus it doesn't break anything.

PR #58, opened 2025-10-09
Fixed Car bouncing, updated textures, and fixed car floating

DrewForYou:
NOTE IF THERE IS A MERGE CONFLICT WITH THE CAR PREFAB TAKE WORKING! changed the floating cars position in both prefab and scene so this should not affect the levels

Added in textures to all the designer levels so they look better. added an invisible roof to all the levels to combat the issue with car bouncing, and edited the cars to stop floating


TheBeesSneeze (approved):
fear this may conflict with the other PRs that edit the scenes but pop off

smnoonan (approved):

PR #64, opened 2025-10-10
Ooo 70

JBBateman:
Fixed all of the bugs listed in the jira tickets.


SkyBeal (commented):
image

noting that I did have a guard get stuck still in patrol state after a vision break

SkyBeal (commented):
Most of the bugs look fixed! I did not see the search state get set off though :[

TheBeesSneeze (approved):
looks good

BBurtz (approved):
Looks good there is still a few kinks to work out but you already know about them so I'm not gonna restate them here

PR #65, opened 2025-10-11
OOO-60

SkyBeal:
possession bug fixed!


TheBeesSneeze (approved):

BBurtz (approved):
Looks good, I played through all the levels twice and not a single possession not working, you can possess things through walls but that is a small thing that we can change for a later milestone

PR #67, opened 2025-10-11
OOO-83

JBBateman:
Guards should no longer get stuck in patrol after a vision break.


TheBeesSneeze (approved):
looking good cooking good

SkyBeal (approved):
works for me!

PR #68, opened 2025-10-14
Final Milestone 1 Build MR

CalebtheMann:
1. All Player Spawn Locations are good
2. Flipped around Adam's car in the first level
3. Extended all of the timers for each of the possessables in all levels (not by a lot, but there's no tutorialization, so its mainly for the sake of the prototype)
4. Adjusted Car Speeds in all levels (BUG: Speeds of car are different in project VS build... Potentially a frame rate dependent issue?)
5. Fixed Vending Machines in all levels


smnoonan (approved):

bboster (approved):
approved causes no technical issues

PR #69, opened 2025-10-17
OOO 51

TheBeesSneeze:
Billboarding UI system.

I applied the billboarding to jacobs debug guard text to test it out.

The Billboard UI is working if it:
1. scales up/down depending on distance from player
2. smooths opacity depending on distance from player
3. The opacity should smooth to zero if the UI is on the edge of the screen.
4. Is always facing the player.


bboster (approved):
i trust ur math on this one king
code looks good tho!

Veizyl (approved):
This looks good to me

PR #70, opened 2025-10-20
Ooo 95

TheBeesSneeze:
* New game prompts player "are you sure" ONLY if they have a save fiIe where they have compIeted 1+ IeveI
* Continue button is hidden if there is no save data (same conditions as above)
* New game button clears save data.
* Quit button
* Credits button

image
image
image


bboster (approved):
looks good approved

SkyBeal (approved):
looks good!

PR #71, opened 2025-10-21
OOO 23

JRavioli:
Temporary reworks to the sound system so sound works and attenuation tests can start tomorrow in class along with recordings for any necessary sfx.


JBBateman:
Fmod is having issues recognizing and tracking fmod sound assets. Closing until we find a fix.

PR #72, opened 2025-10-21
Ooo 96

TheBeesSneeze:
Pause menu
image


bboster (approved):
approved looks good

Veizyl (changes requested):
I am noticing some issues with this:

- After pressing quit to hub in a level the pause menu no longer works in the lobby.
- Pausing does not prevent the player from pressing E on a door

bboster:
> I am noticing some issues with this:
> 
> * After pressing quit to hub in a level the pause menu no longer works in the lobby.
> * Pausing does not prevent the player from pressing E on a door

Ooo you right toby fix this

Veizyl (approved):
Looks good now! Great work!

PR #73, opened 2025-10-21
Ooo 69

TheBeesSneeze:
Swap materials when you enter possessable. Not every possessable has a material yet, so I've got some temp ones in there.


JBBateman (approved):
The system seems to work. I tested a few possessables and also tried to break the pause menu along with it and nothing went wrong in-game. The only not I have is that there is a NullReferenceException that triggers on line 148 in InputEvents when you enter a level.

bboster (approved):
approved, looks good. i agree with jacob's assessment

PR #74, opened 2025-10-21
Ooo 93

JBBateman:
Checkpoint system works. Can be tested in Guard Pathing & Detection Scene. To ensure that it works properly between scenes, you can press L while in editor to load a copy of the scene that should reset the LevelManager's checkpoint.

NOTE THAT ALL LEVELS WILL NEED TO HAVE THE PLAYER DELETED AND AN OBJECT PLACED IN THE WORLD AND DRAGGED INTO THE GameManager TO ACT AS PlayerStart. THIS IS BECAUSE THE PLAYER IS NOW SPAWNED WHEN A LEVEL LOADS BY THE GAMEMANAGER.


Geist Heist/Assets/Scripts/Environment/Checkpoint.cs
```diff
@@ -0,0 +1,21 @@
...
+ */
+using UnityEngine;
+
+public class Checkpoint : MonoBehaviour
+{
+    [SerializeField] private Transform spawnLocation;
+
+    private void OnTriggerEnter(Collider other)
+    {
+        if(other.gameObject.TryGetComponent(out ThirdPersonInputHandler handler))
```
TheBeesSneeze: Would we want checkpoints to be triggered by possessed items as well? Like what if the car zips through a checkpoint.
Could we do "if(other.gameObject == PlayerManager.Instance.CurrentObject)" 
(i forget the exact names of variables but u get the idea)

Geist Heist/Assets/Scripts/GameManager.cs
```diff
@@ -35,16 +37,20 @@ public class GameManager : Singleton<GameManager>
...
 
     [Header("Debug")]
     [ReadOnly] public bool IsPaused = false;
 
+    public GameObject Player;
+
     public static Action OnInitialize;
 
-    protected override void Awake()
+    protected override async void Awake()
```
TheBeesSneeze: Maaaaybe try to avoid async. Especially because we dont NEED to wait for any timers or things right now.
Unity is singlethreaded, so InstantiateManagers does not need to be async:
Without async, the entire game will just freeze until they are all instantiated. (I say freeze but obviously it wont take very long).
WITH async, InstantiateManagers will run multithreadedly. the game will continue loading and other scripts Awake() functions (and possibly start funcitons?) will run as well. If any of those Awake functions rely on any manager singletons then we may be cooked.

I've also had issues with it breaking webgl builds in the past :P

TheBeesSneeze: I'm ready to eat my words on a silver platter and get proven wrong btw

Geist Heist/Assets/Scripts/GameManager.cs
```diff
@@ -35,16 +37,20 @@ public class GameManager : Singleton<GameManager>
...
     {
         base.Awake();
```
TheBeesSneeze, line 55: could you add a 
if(this == null): return;

if there's two gamemanagers in the scene, i believe it will spawn all managers anyways. even if the singleton deletes itself.

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/ChaseBehavior.cs
```diff
@@ -60,7 +60,7 @@ public override IEnumerator BehaviorLoop()
     /// <returns></returns>
     private Vector3 GetPlayerLocation()
     {
-        return GameObject.Find("3rd Person Player").transform.position; //REPLACE WITH CENTRALIZED REFERENCE FROM A MANAGER ONCE ABLE.
+        return GameManager.Instance.Player.transform.position; //REPLACE WITH CENTRALIZED REFERENCE FROM A MANAGER ONCE ABLE.
```
TheBeesSneeze: There already was a centralized reference in PlayerManager.Instance.
Sorry about that, you had literally no way of knowing.

TheBeesSneeze: WAIT thought about it more and now wondering if that reference is still going to work... can you maybe delete one of those variables from one of those scripts (or better yet, replace it with a lambda that points => to the other reference)

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/VisionBreakSearchBehavior.cs
```diff
@@ -21,7 +21,7 @@ public class VisionBreakSearchBehavior : GuardMovement
     public override void InitializeBehavior(GameObject selfRef)
     {
         base.InitializeBehavior(selfRef);
-        SearchLocation = GameObject.Find("3rd Person Player").transform.position; //REPLACE WITH A CENTRALIZED REFERENCE TO THE PLAYER WHEN ABLE
+        SearchLocation = GameManager.Instance.Player.transform.position;
```
TheBeesSneeze: Same note as earlier with the PlayerManager player reference

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/VisionCone.cs
```diff
@@ -22,7 +22,7 @@ public class VisionCone : MonoBehaviour
 
     private void OnTriggerEnter(Collider other)
     {
-        if(other.gameObject.name.Equals("3rd Person Player") && hasSeenPlayer == false)
+        if(other.gameObject.Equals(GameManager.Instance.Player) && hasSeenPlayer == false)
```
TheBeesSneeze: Same note as before with the PlayerManager player ref

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/VisionStimulus.cs
```diff
@@ -38,7 +38,7 @@ public class VisionStimulus : Stimulus
 
     private void OnTriggerEnter(Collider other)
     {
-        if (other.gameObject.name.Equals("3rd Person Player") && hasSeenPlayer == false)
+        if (other.gameObject.Equals(GameManager.Instance.Player) && hasSeenPlayer == false)
```
TheBeesSneeze: Same note as before with the PlayerManager player ref
sorry if these are redundant, i personally appreciate it when i get checkboxes with everything tho.

Geist Heist/Assets/Scripts/LevelManager.cs
```diff
@@ -0,0 +1,59 @@
...
+public class LevelManager : Singleton<LevelManager>
+{
+    private int previousLevel = -1;
+
+    [HideInInspector] public Vector3 SpawnLocation;
+
+    protected override void Awake()
+    {
+        base.Awake();
+        DontDestroyOnLoad(this);
```
TheBeesSneeze: Inherit from the class DontDestroyOnLoadSingleton instead.
Once again literally no way you were expected to know about this.

Geist Heist/Assets/Scripts/Environment/Checkpoint.cs
```diff
@@ -0,0 +1,21 @@
...
+ * Contributors:
+ * Creation: 10/21/25
+ * Last Edited: 10/21/25
+ * Summary: Handles in-level checkpoint logic
+ */
+using UnityEngine;
+
+public class Checkpoint : MonoBehaviour
+{
+    [SerializeField] private Transform spawnLocation;
```
TheBeesSneeze, line 12: this is a really little thing, but a gizmo that shows where this spawn location is (without needing to have the gameobject selected in the inspector) would be really cool. totally optional.

TheBeesSneeze (changes requested):
also i found the issue with the guard debug text. its because its start function happens after the guards debug text gets set. i pushed a change to your branch. feeI free to deIete it and throw stones at me if my fix was bad.

Geist Heist/Assets/Scripts/GameManager.cs
JBBateman: It did wind up causing an issue so I applied the change and it fixed it.

Geist Heist/Assets/Scripts/Environment/Checkpoint.cs
JBBateman: Implemented this and it should now work as described

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/ChaseBehavior.cs
JBBateman: I found the reference and replaced some stuff

Geist Heist/Assets/Scripts/GameManager.cs
JBBateman: Added

Geist Heist/Assets/Scripts/LevelManager.cs
JBBateman: Swapped

Geist Heist/Assets/Scripts/Environment/Checkpoint.cs
JBBateman: I'll need to research gizmos a little since I'm not quite as familiar with them, so I won't do it for this pr, but I'll keep it in mind for an improvements round.

TheBeesSneeze (approved):
my thermostat actually exploded because of how fire this pr was

bboster (approved):
approved

PR #75, opened 2025-10-21
OOO-99

SkyBeal:
WANT FEEDBACK

camera starts on player in every scene, camera does not clip behind walls IF they are box colliders

tested in Player Possession Third Person, also tested in many levels with a change (not included in this PR) of making the walls have box colliders


bboster (approved):
looks good, continue to research how to do it without box collider but this is fine for now!

jtkelly18 (approved):

PR #78, opened 2025-10-23
OOO-180

SkyBeal:
made camera less crazy!


bboster (approved):
i think it feels better, and i agree with the use of orbital follow. let's have design mess with this for this milestone and see if any other changes are required. approved for now!

jtkelly18 (commented):
approved *knuckles stamp*

jtkelly18 (approved):

PR #79, opened 2025-10-24
OOO-87

jtkelly18:


Geist Heist/Assets/Scripts/Environment/OptionalCollectableHubDisplay.cs
```diff
@@ -22,21 +22,92 @@ public class OptionalCollectableHubDisplay : MonoBehaviour, IInteractable
...
+                WearableCollectible wearableCollectible = FindAnyObjectByType<WearableCollectible>();
+                if (wearableCollectible != null)
+                {
+                    wearableCollectible.EquipHat(ThisCollectable);
+                    SaveDataManager.Instance.MarkCollectableAsWorn(ThisCollectable);
+
+                    Debug.Log($"{ThisCollectable} equipped successfully!");
+
+                    foreach (Transform child in this.transform)
+                        Destroy(child.gameObject);
```
TheBeesSneeze: Not sure why we're doing this(?)
Also, this is going to end up breaking the button popup when that is added soon, because that popup is childed to an empty transform under interactables.

Geist Heist/Assets/Scripts/Environment/OptionalCollectableHubDisplay.cs
```diff
@@ -22,21 +22,92 @@ public class OptionalCollectableHubDisplay : MonoBehaviour, IInteractable
...
-    public void Start()
+    public void spawnMesh(Collectable collectableToRespawn)
+    {
+        // Only respawn if this OptionalCollectable matches the collectible
+        if (ThisCollectable != collectableToRespawn)
+            return;
+        GameObject meshPrefab = Registry.GetMesh(ThisCollectable);
+        // Clear any existing children (optional)
+        foreach (Transform child in transform)
+            DestroyImmediate(child.gameObject);
```
TheBeesSneeze: same thing again

Geist Heist/Assets/Scripts/Player/Movement/InputEvents.cs
```diff
@@ -123,7 +123,7 @@ private void FixedUpdate()
         else MoveNotHeld.Invoke();
         //if (JumpPressed) JumpHeld.Invoke();
         if (ActionPressed) ActionHeld.Invoke();
-        else ActionNotHeld.Invoke();
+        //else ActionNotHeld.Invoke();
```
TheBeesSneeze: can we uncomment this? this will break things.

Geist Heist/Assets/Scripts/Environment/OptionalCollectableHubDisplay.cs
TheBeesSneeze: Okay looked at more code and i get it now. Do you think that instead of destroying the gameobject just to make a new one, we can do a GetComponentInChildren<MeshFilter> and swap out the mesh like that? only spawn in a new mesh renderer if there is no mesh childed.

Geist Heist/Assets/Scripts/Player/WearableCollectible.cs
```diff
@@ -0,0 +1,133 @@
...
+        // Get the correct mesh *each time*
+        GameObject meshPrefab = Registry.GetMesh(currentHat);
+        if (meshPrefab == null)
+        {
+            Debug.LogWarning($"No mesh prefab found for {currentHat}.");
+            return;
+        }
+
+        // Find the Wearable node
+        GameObject wearableNode = GameObject.Find("Wearable");
```
TheBeesSneeze, line 52: we should make this an inspector variable, or at least cache it. GameObject.Find can cause significant lag issues. if anything, worst case can we use GameObject.FindObjectOfType ?

Geist Heist/Assets/Scripts/Player/WearableCollectible.cs
```diff
@@ -0,0 +1,133 @@
...
+        // Destroy any existing hat
+        Transform wearableTransform = wearableNode.transform;
+        int childCount = wearableTransform.childCount;
+        Transform[] children = new Transform[childCount];
+        for (int i = 0; i < childCount; i++)
+            children[i] = wearableTransform.GetChild(i);
+
+        foreach (Transform child in children)
+        {
+            DestroyImmediate(child.gameObject); // or Destroy(child.gameObject) at runtime
```
TheBeesSneeze, line 68: same thing, can we not just replace the mesh in the mesh filter?

Geist Heist/Assets/Scripts/Player/WearableCollectible.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    public void ReplaceHat(Collectable previousHat)
+    {
+        // Move the current hat back to hub instead of destroying
+        if (previousHat != Collectable.None)
+        {
+            // Find all OptionalCollectable objects in the scene
+            var allHubDisplays = FindObjectsByType<OptionalCollectableHubDisplay>(FindObjectsSortMode.None);
```
TheBeesSneeze, line 88: can we cache the results of this FindObjectsByType?

Geist Heist/Assets/Scripts/Environment/OptionalCollectableHubDisplay.cs
```diff
@@ -45,4 +115,24 @@ private void UpdateVisibility()
...
+
+    #region Debug Tools
+#if UNITY_EDITOR
+
+    [Button("Preview Collectable")]
+    private void PreviewHat()
+    {
+        if (Registry == null)
+        {
+            Registry = Resources.Load<CollectableRegistry>("CollectableRegistry");
```
TheBeesSneeze: "CollectableRegistry" is a hard coded string thats used in a lot of different scripts. could we make it a public const variable on the registry script to at least have it consolidated in one spot?

Geist Heist/Assets/Scripts/Player/WearableCollectible.cs
```diff
@@ -0,0 +1,132 @@
...
+{
+    public CollectableRegistry Registry;
+    private Collectable currentHat;
+    private Collectable previousHat;
+
+    private void OnValidate()
+    {
+#if UNITY_EDITOR
+        if (Registry == null)
+            Registry = Resources.Load<CollectableRegistry>("CollectableRegistry");
```
TheBeesSneeze: this is soooooo nitpicky but why are we doing this?
i know some code runs in the editor, but this same null check is ran in that code as well.

TheBeesSneeze (changes requested):
ok sorry to be a nitpick-nelly over here. code works and i couldnt find any bugs with it, and thats the important part.

Geist Heist/Assets/Scripts/Environment/OptionalCollectableHubDisplay.cs
```diff
@@ -36,103 +49,90 @@ public void Interact()
...
         {
             Debug.LogWarning($"[{name}] No CollectableModel to respawn for {ThisCollectable}");
             return;
         }
 
-        // Instantiate the mesh
-        GameObject meshInstance = Instantiate(meshPrefab, transform);
+        // Clear existing children to avoid duplicates
+        foreach (Transform child in transform)
+            DestroyImmediate(child.gameObject);
```
TheBeesSneeze: Is it possible that we dont destroy all children? This will break the button prompt once that gets added.
I feel like it would be a lot easier to just grab the MeshRenderer component that already exists from the children. Or maybe make a serialized variable to set in the inspector.

TheBeesSneeze (changes requested):
all these changes are huuuuge. we're very close

Geist Heist/Assets/Scripts/Environment/OptionalCollectableHubDisplay.cs
bboster: ^^

TheBeesSneeze (approved):
banger!!!

bboster (approved):
approved & merged

PR #80, opened 2025-10-24
Ooo 164

JBBateman:
Added ability for guards to be stationary. There is a prefab that can be used that defaults to StationaryBehavior but a normal guard can also be swapped over if necessary. Instructions for setup are contained in the documentation for the guards: https://docs.google.com/document/d/1mQfjLFJi-LJfnkFWXgoVGwykWNVJ60l2XhE2iy8Tol4/edit?pli=1&tab=t.0


bboster (approved):
approved, but please review conflicts

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/Behavior.cs
```diff
@@ -11,25 +11,27 @@
...
 
 public class Behavior : ScriptableObject
 {
     #region Variable Declarations
 
+    [Header("Base Behavior Values")]
+    [SerializeField] private bool showBaseValues;
     [Tooltip("The speed the guard will travel at while performing this behavior")]
-    [SerializeField] private float speed;
+    [SerializeField, ShowIf("showBaseValues")] private float speed;
```
TheBeesSneeze, line 23: I'm confused about why we _added_ another variable that just controls if a variable is shown. Could we make it fit in a [Foldout] attribute instead if we want to hide it instead?

TheBeesSneeze (approved):
ts so flames a fire alarm went off

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/Behavior.cs
JBBateman: Probably I'm just not very familiar with NaughtAttributes.

PR #82, opened 2025-10-27
Ooo 106

DrewForYou:
Level 1 greybox and other small tweaks with values, the guard behavior speeds being the only one I can think of outside of the scene. Overall just want this to be done so things can move forward, then we can work on it more as we get the proper assets in.


bboster:
I can't see the level right now on my machine. When I get home I will see if it's just an issue on my mac. Will let you know.

bboster (approved):
looks good

smnoonan (approved):

PR #83, opened 2025-10-27
OOO-162

JBBateman:
- Added a buffer timer between the player exiting the guard's vision cone and the guard entering vision break state
- Updated some scripts which were lacking headers
- Condensed some repeat code in the behaviors


bboster (approved):
approved

BBurtz (approved):
Everything looks good on my end!

PR #85, opened 2025-10-28
OOO-157

SkyBeal:
toy car should move at consistent speed in both editor and build


TheBeesSneeze (changes requested):
I hate to say it bro but i dont think this is going to change anything, besides you're running the same code in update...
You should look into the forcemode documentation: https://docs.unity3d.com/6000.2/Documentation/ScriptReference/ForceMode.html

the way you had it set up before was fine to my pure eyes.

I think ForceMode.VelocityChange might be your best bet here if you want it to be similar to how it was. 
if you wanna get crazy and apply a constant force over multiple frames (and completely change physics), then ForceMode.Force and ForceMode.Acceleration may be your friends

TheBeesSneeze (approved):
implementation is scuffed but ill admit when im wrong.

bboster (approved):
approved

PR #87, opened 2025-10-28
Ooo 23

JRavioli:
FMOD is not working on anybody else's end because of a directory issue due to how the project was setup and modified after FMOD was imported, so FMOD needs to be completely reinstalled and setup from scratch in a new ticket in order to attempt to get it to work.


bboster (approved):
just removing files, looks good

JBBateman (approved):

PR #92, opened 2025-10-28
Ooo 98

TheBeesSneeze:
Finished possessable timer cooldown guys for individual objects.
image


bboster (approved):
approved

JBBateman (approved):
Timer works. I would keep an eye out for bugs on the vending machine since it's still broken in this branch overall and as such I couldn't test it specifically.

PR #93, opened 2025-10-28
OOO-94

jtkelly18:


TheBeesSneeze:
Also contains ooo-88
Has level transitions. no animations are playing but they can be easily added. no assets were given.

bboster (changes requested):
a little confused on this one..
the player is supposed to be able to walk through a trigger (or oncollisionenter) and have a title card pop up. after this title card pops up there is a coroutine that waits 5 seconds (or however long) and then the card fades and they are transitioned to the level. is this functionality anywhere? i only see a confirmation pop up...

bboster (approved):
approved for now but this may be changed in the future?

SkyBeal (approved):
worked on my end!

PR #94, opened 2025-10-28
OOO-159

SkyBeal:
car now turns with A and D
all possessables now detect if exit points collide with anything before ghost exits possessable
possessables allow for multiple exit points
start of a new gizmos script (that can be expanded)


bboster (approved):
approved

JBBateman (approved):
Works for me

PR #98, opened 2025-10-30
Ooo 166

JBBateman:
- Guard will now continue chasing the player if they possess an object while being chased
- Guard will start to chase a player who is moving in a possessable object
- Guard will now chase a player who uses an action while in a possessable object
- Fixed an error that cause navmesh to back the patrol point visual indicators as walkable surfaces


bboster (approved):
good

BBurtz (approved):
Works like intended

PR #110, opened 2025-11-03
oOo 226

TheBeesSneeze:
Got rid of cooldown manager bc i thought it was the problem (it wasnt).
The problem is EventSystem was missing from the scene. Added it to GameManager so we dont get this problem again.


JBBateman (approved):
Worked both in engine and in a build in the vertical slice scene.

bboster (approved):
good

PR #111, opened 2025-11-03
OOO-227

JBBateman:
Toy car should no longer launch a guard when ramming into one


bboster (approved):
works in build, approved

SkyBeal (approved):
works on my end!

PR #112, opened 2025-11-03
Ooo 212

BBurtz:
This fixes both 212 and 224 works in build


bboster (approved):
approved, doesn't work when scene reloads though. this will hopefully be fixed by josh's ticket.

jtkelly18 (approved):
Approved assuming the issue with scene reloading will be fixed by my ticket

PR #113, opened 2025-11-03
OOO-225 + OOO-223

SkyBeal:
ghost exit points readded - player now exits beside car.
new prefab added called Car Wall, when used around a level, the car cannot go past it. everything else can.
added default number in car script for rotationrate


smnoonan (approved):

JBBateman (approved):
Works in build and in engine.

bboster (approved):
works great

PR #117, opened 2025-11-04
OOO-234

bellawilsonux:
Added a "How to play" and "Credits" page to the buttons on the title screen. They already had functionality so I just styled them. They should function as intended.


TheBeesSneeze (approved):
yess bella hop on the ui implementation grind

bboster (approved):
good

PR #120, opened 2025-11-04
OOO-165

JBBateman:
Guard now has a stunned behavior. While triggered, the guard should freeze for a couple seconds, then return to its default state (either stationary or patrol).

How to test:
- Hit it with a toy car, the guard should enter stunned state
- Hit it with a can (admittedly this one is quite hard to pull off), the guard should enter stunned state


jtkelly18 (approved):
Works in editor and in build, stunning the guard feels pretty satisfying. Fun and requires a bit of skill to pull off good job 👍

bboster (approved):
i tried very hard to hit a guard but kept failing lmao

build works, code looks good, doesn't break.

PR #121, opened 2025-11-05
OOO-235

jtkelly18:
1. The layers included in the raycast in the ThirdPersonInputHandler script are specified in editor through a drop down. This did not include “obstacles” so walls set to layer “obstacles” rather than “default” were not included in the raycast that checks for walls between player and interactable objects. Adding obstacles to the drop down fixed this.

2. Geometry for raycast needed some tweaks, see ThirdPersonInputHandler,137.


SkyBeal (approved):
works on my end!

bboster (approved):
approved but won't merge until merge conflicts are resolved

PR #122, opened 2025-11-05
OOO-219

JBBateman:
Fix for the guards seeing the player through walls.

To test, make sure that the walls have the collision layer 'Obstacles' so that the raycast will be blocked by them.


bboster (approved):
looks good, builds fine

BBurtz (approved):
Tested in build and works just fine

PR #126, opened 2025-11-05
OOO-229

SkyBeal:
car bounces! just added a physics material


bboster (approved):
good, works in build

Veizyl (approved):
I tested this in engine and in build and it works.

PR #127, opened 2025-11-06
OoO 206

TheBeesSneeze:
Brings back fix for pause menu 
Also brought back vending machine assets.
Main reason for this PR is implementing level transition cards.


bboster (approved):
testing in build, looks great

BBurtz (approved):
Tested in build and everything seems to be working!
For the transition from the title screen to the hub is the splash screen supposed to be the art history splash?

PR #128, opened 2025-11-06
OOo 241

TheBeesSneeze:
Added player rotation
Added player hover bob
Added fade to black from main menu
Replaced credits screen
Added ollie model


BBurtz (approved):
Good and works in build (after a lot of fixes)

bboster (approved):
good

PR #129, opened 2025-11-06
OOO-237

breyes09:
Player can now move in GlobeInside scene, GlobeInside now has tether and possessables appear in it after they are collected in the Art Exhibit level, and temporary door was added to GlobeInside so player can return to hub


smnoonan (approved):

bboster (approved):
approved

PR #133, opened 2025-11-06
OOO-239

Veizyl:
The charge meter for the car now resets after 0.1 seconds. I found that doing it instantly messes with the speed of the car. Let me know if you would like this to be something that designers can change. I can make the time a serialized field instead but I believe the intention is for it to be almost instantaneous of a change so I figured it wasn't needed.


bboster (approved):
gets the point across

TheBeesSneeze (changes requested):
Update the chargeLossRate variable instead. hard-setting the currentStrength will mess with the UI.

TheBeesSneeze (approved):
this pr so fire i could cook a marshmellow on it

PR #136, opened 2025-11-07
Clutter pass

Afederspiel:
Added ropes, vases, carpets, and armchairs in my section


smnoonan (approved):

JBBateman (approved):
Looks like a more filled out level and doesn't break anything. Left a note in the design github updates channel on a likely unintended path through the level.

PR #142, opened 2025-11-09
Added temp sfx for vertical slice

JRavioli:


SkyBeal (changes requested):
audio managers are not currently in scene and cause a bunch of errors: with the managers in, the sound works! Just make sure to pull from main and then add the managers wherever you need sound. (later we can spawn this in with Game Manager)

other things to note for later:
step sounds are pretty quiet while the whistle is super loud
sound does not stop when paused

bboster (changes requested):
what they said

PR #143, opened 2025-11-10
Ooo 251

JBBateman:
Fix for checkpoint system. Player should no longer re-enter a level after beating it and spawn at the final checkpoint.


SkyBeal (approved):
works in build!

bboster (approved):
works in build

PR #149, opened 2025-11-10
OOO-256

JBBateman:
Stationary guards should properly return to their station after searching an area.


bboster (approved):
good

breyes09 (approved):
Tested in engine, stationary guards do now return to their original position, everything looks good

PR #151, opened 2025-11-13
OOO-208

JBBateman:
Improvements to the guard's inspector layout:

- Variables should be groups better in the GuardController component
- Behavior ScriptableObjects inspector interface moderately improved
- Behavior ScriptableObjects should have programming values moved into a foldout
- Currently running and default behaviors should be expandable on the instances of the guards in the level, allowed viewing of the variables of the behavior instances during rutime.


Veizyl (approved):
This looks good to me. The inspector for guards looks more organized for sure.

bboster (approved):
good

PR #153, opened 2025-11-14
ooO 272

TheBeesSneeze:
Added new UI that appears when you get a checkpoint


JBBateman (approved):
Works in engine and in build. I will not that the UI element pops up when getting caught by a guard and reset to a checkpoint, but that doesn't feel like a big deal.

bboster (approved):
merged

PR #156, opened 2025-11-17
Ooo 262

BBurtz:
PA System is in, effectively just the same as dialogue system currently but there will have to be some changes done when making the audio section working with joey.

To test Just drag a PA trigger, in prefabs and UI folders, into the main scene (I didn't save the changes so that it wouldn't cause any merge conflicts) the PA box should be bright red just to make sure that its triggering the correct box instead of just using the Dialogue system's text box.


Geist Heist/Assets/Scripts/UI/PATrigger.cs
```diff
@@ -0,0 +1,48 @@
...
+            DailougeManager.Instance.PATextbox.text += Text.Substring(temp, 1);
+            temp++;
+            yield return new WaitForSeconds(.05f);
+        }
+        yield return new WaitForSeconds(stayLength); //this will be replaced with the end of the audio clip eventually
+        clearBox();
+    }
+
+    //this is just in case we have to have it called somewhere else for the audio clip ending when that gets implemented
+    private void clearBox()
```
TheBeesSneeze, line 44: Make this public?

TheBeesSneeze (approved):
hell yeah looks good

bboster (commented):
code looks good, i will build and determine approval around 7-8pm this evening

also lmao i never realized dialogue was called dailouge, is there a reason for this?

bboster (approved):
builds fine

PR #157, opened 2025-11-17
OOO-260

SkyBeal:
player cannot leave toy car while it's moving

timer is paused when toy car is moving


bboster (commented):
code looks good, i will build and determine approval around 7-8pm this evening

bboster (approved):
works in build, approved

BBurtz (approved):
Looks good, this is a good fix for the softlocks, good job

PR #159, opened 2025-11-18
OOO-264

Veizyl:
Ollie now uses raycasts to determine whether or not they can float up something (stairs in this case)


bboster (commented):
waiting to approve this one until we see updates per caleb's request during class 11/18

Veizyl:
This was tested with the assumption that OOO-318-2 gets merged with the player speed being updated to 8.5 instead of 7. 
This works with slope transition smooth at 0.5 and slope modifier at 1 AND the player speed at 8.5.  If speed is not updated then those values will have to be changed.
Z position is locked while on slope otherwise the player moves unreasonably fast horizontally across the ramp. This does take away some control from the player but it is the best solution that I have found.

JBBateman (approved):
Works in engine and in build.

bboster (approved):
approved

PR #161, opened 2025-11-18
OOO-290

SkyBeal:
hat particles! 
changed test prefab to have particles so it should automatically be on all collectibles


jtkelly18 (approved):
Looks good, approved

bboster (approved):
looks great!

PR #162, opened 2025-11-18
Ooo 316

TheBeesSneeze:
New flavor text interactables. 
UI is temporary. I don't think it will be hooked into PA system later. Made a new ticket for implementing improved UI (see OOO-339). Going to talk to bella soon.

Designers can set the dialogue to only appear based on certain conditions (levels completed / not completed, wearing specific hat, etc). This wasn't really asked for, but I had fun.
image

I also noticed a bug with the hat system, where it wouldn't support not wearing a hat.


bboster (commented):
code looks good, i will build & test later this evening

BBurtz (approved):
Looks good

bboster (approved):
builds fine looks good

PR #163, opened 2025-11-19
OOO_318

breyes09:
Balancing Changes (Can be viewed in the Guards Maze Testing scene in the Testing folder (In the Scenes folder)):

- Player speed raised
- Guard chase speed raised (Cannot currently change turn speed of guards, will come back and make more changes when that's accessible)
- Raised toy car max speed and turn speed
- Raised vending machine launch distance so it can launch cans instead of them just being launched directly in front of the vending machine


Sulphide2 (changes requested):
The NavMesh for Level 1 was deleted entirely so it doesn't work anymore, that should be added back in. Otherwise it's fine.

bboster (approved):
in build the guards do not work, please review

Veizyl (changes requested):
Everything seems to work fine in the testing scene but in Level 1 Art Exhibit (in engine or build) some of the guards are getting stuck on something.

PR #164, opened 2025-11-19
OOO-289

SkyBeal:
particles for guards when in chase mode


bboster (commented):
code looks good, i will build and test later this evening

JBBateman (approved):
Works on my end.

bboster (changes requested):
build fails, please fix

bboster (approved):
builds good!

PR #165, opened 2025-11-20
Rotation Speed Can Be Changed

JBBateman:
Rotation speed can be changed by altering Rotation Speed and Rotation Acceleration on each behavior.


jtkelly18 (approved):
Works

bboster (approved):
works, had to comment out "using UnityEditor.ShaderGraph.Internal" in the GuardController.cs file as it was causing build errors

PR #167, opened 2025-11-20
Ooo 332

smnoonan:
Re-scaled outer and inner walls, grey boxed interaction in first room.


SlicedBreadStudios:
I'm assuming this is the parlor room 1 scene. 

- Can you increase the wall height to what it might be in the final so we can see how it plays with walls before moving on? 
- Can you move the player back to the start when you load the scene
- This feels better than the version from this morning but still too big. Play through it with walls and ask for feedback before deciding to make it bigger or smaller.

bboster (approved):
good

SlicedBreadStudios (approved):
This is fine based on the videos in discord

PR #168, opened 2025-11-21
Ooo 97

BBurtz:
Debug console is in
Currently only dragged into the level 1 art exhibit (VS) scene but you can drag it in in any scene and it will work (has to be dragged in to each scene so we can pick which scenes do and don't have the debug console a lot easier)
Use help command to see all the commands

Some caveats:
Freeze command is currently not implemented due to me needing to find a time to work with Jacob on a freeze state for the guards for this command
camera freaks out when you walk though walls in no clip
There is a bug with detached camera when the player is seen where the guard will try and attack the free cam'd player instead of attacking the player in front of them (not too big of a deal but I will talk with Jacob about this


Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+        else
+        {
+            Debug.Log("No Console Exists");
+        }
+    }
+
+    public void CallFunction()
+    {
+        string Command = inputs.text;
```
TheBeesSneeze: super nitpicky, why not just have
string Command = inputs.text.ToLower();
so you dont have to call it in every if statement?

Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
```diff
@@ -0,0 +1,162 @@
...
+            if (cameraToggle)
+            {
+                PlayerManager.Instance.PossessFreecam(FreeCamInstance.gameObject.GetComponent<PossessableObject>());
+            }
+            else
+            {
+                PlayerManager.Instance.PossessGhost(FreeCamInstance.gameObject.GetComponent<PossessableObject>());
+            }
+        }
+        else if(Command.ToLower().Substring(0,2) == "ls")
```
TheBeesSneeze: this command throws errors if you call it without putting anything at the end of it. 
tbf you could say thats the users fault ¯\_(ツ)_/¯

Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+        else if(Command.ToLower() == "help")
+        {
+            TextArea.text = TextArea.text + "\n" + Command + "\nNo Clip: nc \nGod Mode: god \nDetatch Camera: dc \nFreeze Guards: freeze " +
+                "\nLoad Scene: ls <Scene Name/Scene Index> \nSpawn Item on camera: spawn <Item Name/Item Index>";
+            Debug.Log("Help");
+        }
+        else if(Command.Length > 4)
+        {
+            if (Command.ToLower().Substring(0, 5) == "spawn")
```
TheBeesSneeze: this command also throws errors if you dont have any input after the command.

Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
TheBeesSneeze: for clarity, i typed "ls" and it gave an error

TheBeesSneeze: unrelated issue, i had a lot of trouble getting this command to work because i thought the help text said "is" (with a capital i. I think its the font)
could you change the command to be "scene" instead of "ls"? you could probably keep the "ls" functionaIity and just add another command that does the exact same thing.

Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
```diff
@@ -0,0 +1,162 @@
+using System;
```
TheBeesSneeze, line 8: dont forget your file headers 🤓👆

Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    public void CallFunction()
+    {
+        string Command = inputs.text;
+
+        if(Command.ToLower() == "nc")
+        {
+            Debug.Log("no clip");
```
TheBeesSneeze: this is incredibly nitpicky, especially for debug code, so only do this if you want, but this long if-else chain can be a little hard to parse. do you think you could make individual functions for each Debug Command?

TheBeesSneeze: and only do this if u have time. thanksgiving IS right around the corner.

Geist Heist/Assets/Scripts/UI/Debug Console/FreeCamMode.cs
```diff
@@ -0,0 +1,125 @@
+using GuardUtilities;
```
TheBeesSneeze, line 1: dont forget your file headers! 🤓👆

TheBeesSneeze (changes requested):
HUGE!!! Ty so much for your work. i have a million requests but at the end of the day your shit works and it works good. tysm soldier 💪

Some requests i have:
* could you make the debug console spawn in at runtime from gamemanager? so that way we dont have to drag it into every scene
* could you write down somewhere in the documentation that ` opens the debug menu?

Some smaller QOL things i think would be cool (you dont *have* to do ofc):
* could you make the default text in the textbox  say something like "Enter command (enter 'help' for list of commands)". i do appreciate that it says "use help for all commands" tho, thats pretty cool.
* i could not figure out how to use the load scene function without looking at the code. it was kind of unintuitive. 
* i do think it would be cool if the player was teleported to the cameras position when exiting freecam mode

TheBeesSneeze:
one other small note, it doesnt matter much now that its in and it works, but for future projects, unity does come with its own input field component. i think you might have overengineered your input system.
image

Geist Heist/Assets/Scripts/UI/Debug Console/DebugConsole.cs
BBurtz: This was so when I would append the message onto the next line that would show up on the console it would be in the same capitalization that the user put in but it genuinely doesn't matter

TheBeesSneeze (approved):
HUGE!!! tysm goat

bboster (approved):
awesome work!

PR #171, opened 2025-11-21
OOO-318-2

breyes09:
Balancing changes from the original OOO-318 branch were copied over to avoid the navmesh issues in the vs level in the original OOO-318 branch. Guard and player speeds have been raised, vending machine shooting is fixed, and toy car charges and rotates faster and moves quicker at max speed


bboster (approved):
looks good builds fine

Veizyl (approved):
Works in build and engine

PR #173, opened 2025-11-23
OOO-284

JBBateman:
Added temp guard animations.


breyes09 (approved):
Looks good!

bboster (approved):
guards do get stuck on occasion, but since this is temp we can merge for now

PR #174, opened 2025-11-24
OOO-286

Veizyl:
Ollie now has particles while moving.


JBBateman (approved):
Works in build and in engine.

breyes09 (changes requested):
Particles no longer appear after the player possesses a vase. Particle density feels fine to me, though it would be nice if they disappeared a bit quicker.

bboster (approved):
i cannot replicate bug breyden mentioned, i will merge for now and we will keep this in mind...

bboster (approved):

breyes09 (approved):

PR #175, opened 2025-11-24
OOO-287

Veizyl:
The possessables now have particle effects while being possessed.


JBBateman (changes requested):
There are a few errors that print to the console that aren't present in other branches as far as I can tell. If they aren't an actual issue I can come back and change this to an approval.

Errors:
Not allowed to access vertices on mesh - This is on the vase possessable's model
UnassignedReferenceException var cameraAnchor of PossessableObject has not been assigned - this seemed to be on each vase and prevented the camera from transitioning over to the vase, leaving it where the player was when they possess the vase

JBBateman (approved):

bboster (approved):

PR #176, opened 2025-11-24
Ooo 285

jtkelly18:
Smoke particle fx for guard chase state.


Geist Heist/Assets/Scripts/Guards/Pathing & Detection/GuardController.cs
```diff
@@ -124,9 +125,11 @@ private void Update()
 
         if (currentBehavior.StateName == GuardStates.chase)
         {
-            if (particleSystem != null && !particleSystem.isPlaying)
+            if (dustParticles != null && !dustParticles.GetComponentInChildren<ParticleSystem>().isPlaying)
```
SkyBeal: You're testing if dust particles are null but not if the other particles are before playing them

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/GuardController.cs
```diff
@@ -47,6 +47,10 @@ public class GuardController : MonoBehaviour
 
     [Foldout("Programming Values")]
     [SerializeField] private Animator animator;
+    [SerializeField] private GameObject dustParticles;
```
SkyBeal: You can set all of these to be of type ParticleSystem so that you don't have to GetComponent later

SkyBeal (changes requested):
overall the form of the particles look pretty good! they are really hard to see though since our environment is pretty dark. could you bump up the brightness so that they're more visible?

also left a few (hopefully quick) format changes

Geist Heist/Assets/Scripts/Guards/Pathing & Detection/GuardController.cs
```diff
@@ -154,9 +169,11 @@ private void Update()
         }
         else
         {
-            if (particleSystem != null && particleSystem.isPlaying)
+            if (dustParticles != null && dustParticles.isPlaying)
```
SkyBeal, line 171: one last area to put the extra null check I promise

SkyBeal (approved):
looking really good! there's one last area that needs an extra null check but I trust it'll take like 2 seconds

bboster (approved):

PR #182, opened 2025-11-25
OOO-268

JBBateman:
Vision cone renders and should deform around any objects labeled 'Obstacle'. Still a little with collision detection and casts at a diagonal.


Geist Heist/Assets/Scripts/Guards/Pathing & Detection/VisionStimulus.cs
```diff
@@ -50,6 +66,11 @@ private void Awake()
...
     }
 
+    private void Start()
+    {
+        //GenerateVisionMesh(); //THIS FUNCTION IS EVIL RIGHT NOW
+    }
+
     private void OnValidate()
     {
         SyncLightToCollider();
```
jtkelly18, line 76: This line could be moved to Update() to have the flashlight beam better sync with the vision cone during guard animations.

jtkelly18 (approved):

JBBateman:
Noticed that the cone sometimes renders off to the side if a raycast collides with a wall. Probably won't be fixed today, so I'll put it on my list for after break.

bboster (approved):
good, since you already have a bug ticket

PR #183, opened 2025-11-25
Search Behavior Fix

JBBateman:
Stationary guards should properly return to their post.


bboster (approved):
good

breyes09 (approved):
Functions correctly

PR #184, opened 2025-11-25
OOO-319 hub clutter progress

Afederspiel:
added lighting and props, needs to fix flavor text interactables


bboster (changes requested):
i love it so far! i just have a few notes that we can maybe work on Monday.

- can we make the text in the hub sound a bit more museum like? i would rather it feel like you're reading from the museums pov rather than ollies

- can we allow the player to click and read the text again after they read it once? may be something toby needs to work on next milestone

JBBateman:
Stair movement is broken in this branch, so I'm holding off on approving this until it gets looked at by Sky Turner.

JBBateman (approved):
Fixed the build error and it works on my end. Note that the ramp and upper level are currently broken, I believe this is being looked into by Sky Turner in another branch.

bboster (approved):
none of the interactables work but this is a known bug so merging

PR #185, opened 2025-11-25
Ooo 347

Sulphide2:


jtkelly18 (approved):
Only the stationary gaurds are replaced with the new prefab but other than that everything seems to be working as intended

bboster (approved):

PR #187, opened 2025-11-25
Ooo 352 [OLD]

smnoonan:
Adding props to red area of level, textures on 3D models 
- Textures on paintings are placeholder 
- Don't yet have modern benches / trashcans


SlicedBreadStudios (approved):
It runs incredibly laggy but it does run. 
There are changes I would like to see made when we come back on monday including:
- Using a variety of paintings
- Making sure the vending machine doesn't feel weirdly scaled in comparison to things
- Changing some sections to make more sense (why is there a bathroom and bedroom area when in theory those will be other wings) 

For the time being this is sufficiently cluttered

bboster (approved):
good

PR #189, opened 2025-11-26
OOO-358

CalebtheMann:
Lots of Bugs, very much just "thrown together" for people to play with over break with family. Need to talk with Brooke about merging and how this was made


DrewForYou (approved):
looks good to me but realize I forgot to replace the greybox of the ropes with ropes

JBBateman (approved):
Works for me

PR #191, opened 2025-12-02
Ooo 361

TheBeesSneeze:
New script for hover button sounds :)


JBBateman:
Just leaving a comment to remind myself that this MR is having some bugfixing done on it

JBBateman (approved):
Flavor text throws a bunch of null refs in the hub scene, so if that shouldn't be happening I would go back and throw in what it needs in order to work. Other than that it looks good.

bboster (approved):
good

PR #193, opened 2025-12-02
OOO-327

MadelineLeif:
Adds the hat collectibles: jester hat, newsboy cap, and cowboy hat to the models folder.


JBBateman (approved):
Works for me.

jtkelly18 (approved):
Models all look good I am foreseeing changes to the collectible system be required to account for moving hats to specific positions during preview and saving those coordinates to the collectible registry hats don't float 3 feet above ollie's head.

PR #194, opened 2025-12-02
OOO-369

Sulphide2:
Guards mostly added properly


bboster (approved):
good

CalebtheMann (approved):
Approved, but one guard isn't hooked up correctly with his patrol points

PR #196, opened 2025-12-02
OOO-356

jtkelly18:
Added animations for the collectibles in level. This includes changes to the TestCollectible Prefab to add animations as well as a new child for the model itself. Additionally made changes to the WearableCollectible script to address being unable to see hats in level as well as some debug tools for manually equipping hats.


bboster (approved):
looks good

PR #199, opened 2025-12-02
OOO-355

SkyBeal:
added fmodevents, audiomanager, and musicmanager to the initialize function in game manager
adjusted each of these to inherit from singleton<T>


JBBateman (approved):
Works

bboster (approved):

PR #200, opened 2025-12-02
OOO-372

Veizyl:
Readded the movement code that allows Ollie to go up and down slopes. Slopes aren't actually in this branch to prevent merge conflicts with OOO-319 but it works I swear


JBBateman:
I think that this pull request should be closed as the slope movement seems to have been recovered with our working fix.

PR #201, opened 2025-12-02
Ooo 362

DrewForYou:


SlicedBreadStudios (changes requested):
You did the walls properly, but there are changes I'd like to see to your level

The paintings on the wall feel 

- too small
- as though they are floating in front of the wall
- like they are all at the same height leaving a lot of blank space higher up

The windows and curtains feel small, they can stand to take up more space
The book cases are empty. There should be prefab with books made from another section that you could use. 
I know the carpets were to draw the player's eyes to the corner but as they stand they feel out of place and awkward.

SlicedBreadStudios (changes requested):
Only things I've got for you is a request to make the paintings bigger to make them take up space while using less of them. I don't want to exhaust all the paintings we have so early in the level

SlicedBreadStudios (approved):
This is good

PR #202, opened 2025-12-02
OOO-374

breyes09:
Added dialogue triggers into first level, then put the triggers into a prefab named dialoguetriggersiterativelevel in the iterativebuildparts folder within the prefabs folder to avoid merge conflicts, the prefab of the can be correctly placed in level by making sure PATriggerLevelStart is centered in the first doorframe/archway of the level


CalebtheMann (approved):
Works great! To make sure the positioning is exactly how it's supposed to be, when we place them in the game, open the prefab, copy the parent, then go back to ParlorRoomRed Scene and paste. Then, Delete the Mesh Renderer. There's temporary text in right now, but it functions perfectly fine in engine and in-build. I re-deleted the dialogue in the scene though so that we can make sure the scene remains untouched

bboster (approved):

PR #203, opened 2025-12-02
Working fix backup

bboster:


JBBateman (approved):
Besides the bugs already known I haven't found any others.

PR #204, opened 2025-12-03
New Tether Vase model and textures + Possessable Vase Eyes

Navahkiin:
Added New tether vase model, textures, and the textures for the possesable vase's eyes.


CalebtheMann (approved):
Approved

bboster (approved):
good just adding art

PR #205, opened 2025-12-03
Ooo 363

smnoonan:
- Added to existing clutter of red section
- Decorative walls
- Benches
- Ropes
- Changed suggested areas 
- Fixed paintings and books
- Prefab: Assets -> Prefabs -> IterativeBuildParts -> RedUpdated


SlicedBreadStudios (approved):
I found the prefab and threw it in the scene to look at. This mostly looks good! 

The paintings in your first room look a little warped and there's almost too many of them if that makes sense. I think for something that feels like an art gallery it's okay for there to be big paintings staggered with some space between them. This is fine for this milestone. If you have time to tweak it great, if not, it can be approved anyway.

PR #209, opened 2025-12-03
OOO-376

CalebtheMann:
2nd Pass all stitched, incredibly laggy though, and enemies aren't moving (which may have to deal with poor performance). This is just so that we have a scene with everything in it so lighting can start


SlicedBreadStudios (approved):
Looks good

JBBateman (approved):
Works for me.

PR #212, opened 2025-12-04
OOO-310

JRavioli:
Added sfx for iterative build


JBBateman (approved):
Looks like it works. Sound turns off entirely sometimes but it looks like it's caused by a GameManager null ref that I believe already has a ticket open.

bboster (approved):
good

PR #222, opened 2025-12-04
OOO-391

Veizyl:
Added ramp layermask to raycast to prevent Olllie from climbing vision cones and props.


BBurtz (approved):
I can't climb anything im not supposed to

bboster (approved):
looks good

PR #225, opened 2025-12-04
OOO-353

JBBateman:
The vision cone should render properly. Note that if it phases through an object it is probably because the object is not tagged Obstacles and so the guard can see through it.


bboster (approved):
good

PR #227, opened 2025-12-05
OOO-378

JBBateman:
The guard raycast should now travel in the proper direction and collide with all objects tagged with Obstacles, preventing the guard from detecting the player through walls.


BBurtz (approved):
they didn't see me through walls now!

bboster (approved):
good

PR #228, opened 2025-12-05
OOO-393

JRavioli:


JBBateman (changes requested):
No sound plays in the main menu the first time through. It does play when returning to the main menu.

bboster (approved):
approving

PR #229, opened 2025-12-05
Ooo 404

smnoonan:
- Connects globe back to hub 
- Fixed Ollie's eye texture
- Car's texture won't switch to possessed but it is in the correct spot


JBBateman (approved):
Looks good.

bboster (approved):
works good besides memory leak which will be fixes with new version

PR #234, opened 2025-12-06
OOO-410

JBBateman:
Allowed read/write access on some models that weren't allowing it and throwing errors into the console.


bboster (approved):
good

CalebtheMann (approved):
Got a couple 5-10 second stutters after being caught, but never a crazy long one or a crash. Also, enemies can't see through walls!

PR #236, opened 2025-12-06
OOO-409

JBBateman:
Removes the GuardDebugger script from the guards so that the build doesn't look for it. If any of my MRs will break the game this is the one so make sure to thoroughly test that each guard in the scene is working.


bboster (approved):
they appear to all be working, i tested all the guards i could get to. let's have caleb test as well.

CalebtheMann (approved):
Guard saw through a wall but behaved normally

PR #237, opened 2025-12-06
OOO-412

JBBateman:
Altered a couple of scripts to destroy an object and implemented a function to clean up the mesh generation for the guard vision cones after every frame to try and avoid garbage collector spikes related to memory allocated for that rendering.


bboster (approved):
good

CalebtheMann (approved):
Works BEAUTIFULLY!!

There is this small bug that we shouldnt worry about with one of the guards, but it's kinda unnoticeable and hard to explain, so I'll talk to you in-person about it during Monday Meeting

PR #238, opened 2025-12-08
OOO - 414

CalebtheMann:
Design Fixes:
1. Kiosk is now the door for the Hub
2. All Hats now work (in globe and FINAL PARLOUR scene)
3. Fixed Dialogue System (placed in scene, implemented the three voicelines and subtitles)
4. Updated the Hats to match the environment
5. Updated Main Menu to say "Demo"
6. Possessables have the correct materials
7. Fixed Final Hat Collider Issue
8. Can no longer leave Hub

Banged my head up against a wall way too many times making this and finding more and more pre-existing bugs/issues. Most notable ones:
1. Guards can just see through certain walls, even if they are marked Obstacles.
2. Guards can catch player at seemingly random times if Player enters their vision (Most consistently happens if you're spotted in the car during the final part of the level)
3. Guards occasionally enter chase state and never leave
4. Your Hat Selection is not saved when entering the Actual Hub Scene
5. Had an issue with the Collectable that forced me to make a decision: All of them functioning, but the Collectables Collected on the Pause Screen always said "out of 4" instead of "out of 3", or one of them not functioning... Can explain that issue in-person if confusing
6. Literally possible to just run straight through the entire game up until you need a car... Definitely a Design oversight
7. Camera is still an issue in vending machines
8. Entering a Possessable while still moving causes the movement sound to continually play until you re-possess something

None of this is gamebreaking at all, and all of these issues I BELIEVE already existed before this MR if you tested it enough times, but wanted to note them just in case.


bboster (approved):
LOOKS GREAT

JBBateman (approved):
Looks good

PR #244, opened 2026-01-26
SFX Review

jtkelly18:
Review of the scripts for SFX, minimal changes for FMODEvents and MusicManager. Audio Manager needs a bit more attention. Also the static Instance -> instance change in FMODEvents will require a find-and-replace in Music and Audio Manager.


BBurtz (approved):
Fixes some redundancy that could have caused issues down the line

JBBateman (approved):
Looks good

PR #246, opened 2026-01-27
OOO-459-481

TheBeesSneeze:
Manager System Review


BBurtz (approved):
Looks good and the game still works just as planned!

SkyBeal (changes requested):
I'm getting a bunch of errors that prevent me from playing in any scene

Image

SkyBeal (changes requested):
Image

Image

getting new errors now </3

SkyBeal (approved):
Does not cause errors IF you play from the main menu. New game is also not working on my side (sounds like a me problem?). Having to play every time from the main menu might be rough for the designers.

Technically works BUT we might want to ask the designers if this would mess with their flow/come up with a different solution for initialization in individual scenes

SkyBeal (changes requested):
Going back on my review (sorry) because I think it will be really bad if we _have_ to start from the main menu for it to work, especially for testing. These requested changes are also partially on behalf of Toby.

JBBateman:
I'm gonna rig up some of the managers to start in the parlor room as well and see what I can do.

SkyBeal (approved):
Works now on scenes besides main menu!

PR #247, opened 2026-01-27
Ooo 460,468 possessable review and revision

JBBateman:


Geist Heist/Assets/Scripts/Player/Possession/PossessableObject.cs
```diff
@@ -119,7 +119,8 @@ private void OnTriggerEnter(Collider other)
         }
     }
 
-    private void OnTriggerStay(Collider other)
+    //I'm not seeing a reason to keep this function if Enter and Exit are being used, but if there's some reason I'm unaware of then this is fine to stay
```
SkyBeal, line 120: I believe the stay was to handle an edge case for the guard looking at an object (guard vision cone enters -> visible but not in danger material -> player causes chase state without guard vision cone exiting + therefore displays wrong material)

BBurtz (changes requested):
Hey jacob, don't forget to get this fully assigned in class Tuesday

JBBateman:
Changes have been made so this is good to review.

BBurtz (approved):
Looks good

TheBeesSneeze (approved):
A lot of comments were not acted upon, but it does look fine.

PR #250, opened 2026-01-29
Ooo 348

BBurtz:
Added the Trajectory predictor for where the can is going to land when launching from the vending machine
Can now shows the onomatopoeia when it lands and it has a despawn timer after it lands so there isn't a bunch of can clutter if you use the vending machine a lot. 
The can no longer can stun the guard by the guard walking over it and making to move
you can test this in the final parlor room scene


Geist Heist/Assets/Scripts/Guards/GuardStunner.cs
```diff
@@ -6,14 +6,15 @@ public class GuardStunner : MonoBehaviour
     [Tooltip("How much velocity this object must have before it can stun a guard")]
     [SerializeField] private float stunVelocityThreshold;
     [SerializeField] private bool isCar;
+    bool canstun = true;
```
TheBeesSneeze: camelCase

Geist Heist/Assets/Scripts/Player/Possession/VendingObject.cs
```diff
@@ -37,20 +35,18 @@ public class VendingObject : IInputHandler, IInteractable
...
 
     [SerializeField] private PossessableChargeMeterUI chargeMeter;
-    [SerializeField] private ParticleSystem possessableParticle;
 
     private PossessableObject possessableObject;
     private bool hasThrownThisPossession;
     private Coroutine materialCountdownCoroutine;
 
-    private EventInstance canCharge;
+    [SerializeField] TrajectoryPredictor TP;
```
TheBeesSneeze: can we rename these to have more appropriate names (trajectoryPredictor is fine). Also give it the Required tag.

TheBeesSneeze (changes requested):
It seems a little bugged when you hold the button while the vending machine is charging back down (like if you press button right after shooting a can). I'm not sure what should happen tbh, but this doesnt feel right.

Also, the line doesnt connect to the vending machine object. I know this is because it starts at the CanSpawnPoint, but i think that means we need to move CanSpawnPoint. Maybe make it so cans dont collide with the vending machine if thats a problem.

BBurtz:
> It seems a little bugged when you hold the button while the vending machine is charging back down (like if you press button right after shooting a can). I'm not sure what should happen tbh, but this doesnt feel right.
> 
> Also, the line doesnt connect to the vending machine object. I know this is because it starts at the CanSpawnPoint, but i think that means we need to move CanSpawnPoint. Maybe make it so cans dont collide with the vending machine if thats a problem.

I see what you were talking about with the whole can holding down thing not feeling right but I think this was something that was added for the UI to work how it was wanted to if I'm not mistaken, I think I know a way to fix it but I think it would have the UI function a bit differently. This is something that we will have to talk about for the fix for it.

TheBeesSneeze (dismissed):
Looks good, ty soldier

TheBeesSneeze (approved):
Looks Good Again

JBBateman (approved):
Looks good.

PR #254, opened 2026-02-01
OOO-407

JBBateman:
Guards should no longer repeatedly gasp when seeing the player.

A few exceptions exist:
1. When the player is in god mode the guard will repeatedly gasp
2. When the player rapidly triggers swaps between vision break and surprised mode


JBBateman:
Apparently it didn't fix it enough.

PR #256, opened 2026-02-03
Ooo 556

TheBeesSneeze:
Fixed bug where "bonk" showed up multiple times when car hits a wall.
Also added some juice to the Onomatopoeias.


JBBateman (approved):
Looks good

SkyBeal (changes requested):
Can you make an editable variable for the speed of the onomatopoeia showing up? I want it to have a little more "kick"

BBurtz (approved):
Looks good, good call on the change sky

SkyBeal (approved):
works well! we should add the same variable to the can later but for now this works :]

PR #259, opened 2026-02-04
Ooo 592

jtkelly18:
Adds a small runtime key system and a locked-door interactable: key types are listed in an enum in KeyType.cs. KeyManager is a scene‑scoped singleton that stores KeyType values and clears on scene load. KeyItem.cs handles key collection and animations, then calls callKeyManager.Instance.AddKey(keyType).

Doors use LockedDoorInteractable.cs. Designers will have to set the KeyType, hinge corner, pivot angle (90 default), open animation length (0.6s deafult); the script then computes the hinge pivot from the door’s child renderer bounds (with an optional inset, can be set in inspector) and rotates the door around that pivot. Front-corner hinges rotate by -OpenAngle (toward the player), back-corner hinges by +OpenAngle (away). Doors query KeyManager.HasKey(...) before opening and play locked/unlocked SFX(which need to be added).

You can test keys in Scene/Testing/Keys


Geist Heist/Assets/Scripts/Environment/Keys/KeyManager.cs
```diff
@@ -0,0 +1,66 @@
...
+ * Call KeyManager.Instance.AddKey(...) from pickups. Use HasKey(...) to check.
+ * Keys are NOT consumed on use.
+ */
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+public class KeyManager : MonoBehaviour
+{
+    public static KeyManager Instance { get; private set; }
```
TheBeesSneeze: inherit from the Singleton base class

Geist Heist/Assets/Scripts/Environment/Keys/KeyItem.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            transform.position = pos;
+            transform.eulerAngles = rot;
+
+            yield return null;
+        }
+
+        Debug.Log("Confetti explosion goes here???"); //TODO:
+
+        // Go to her...
```
TheBeesSneeze, line 102: if ur gonna copy my homework at least don't make it obvious /j

(idgaf abt reusing the code i think this is fine)

Geist Heist/Assets/Scripts/Environment/Keys/KeyManager.cs
```diff
@@ -0,0 +1,66 @@
...
+    private void Awake()
+    {
+        if (Instance != null && Instance != this)
+        {
+            Destroy(this);
+            return;
+        }
+
+        Instance = this;
+        SceneManager.sceneLoaded += OnSceneLoaded;
```
TheBeesSneeze, line 27: move this to an initialize function similar to other managers?

TheBeesSneeze (approved):
Yo ts is fire. My changes are nitpicky bc it works.
Nowhere to put it in the code, but could you delete the button prompt after the door opens?

JBBateman (changes requested):
The code looks good, but I think it should be edited real quick to use Singleton and initialize like the other managers to maintain consistency.

JBBateman (approved):
Looks good to me

PR #260, opened 2026-02-05
OOO-663

JBBateman:
Guards should not see the player through walls anymore.


TheBeesSneeze (approved):
It works

SkyBeal (approved):
I am so safe behind walls

PR #261, opened 2026-02-05
Ooo 533

BBurtz:
Overhead system is done, It doesn't currently disappear because that will be tied to the length of the voice line that will be played but it is coded to be able to do that and I've talked with joey and he said that it should be possible to do that on his end. Only currently works in the surprise state for the same reason because it will change line from line.
It should be testable for every guard but it works in the parlor room scene for sure.


JBBateman (changes requested):
The system itself works great but the text assets have been completely overwritten and Q now appears over every interactable object at all times.

Geist Heist/Assets/Scripts/Guards/Behaviors/SurprisedBehavior.cs
```diff
@@ -41,6 +41,7 @@ public override IEnumerator BehaviorLoop()
         if (playbackState.Equals(PLAYBACK_STATE.STOPPED))
         {
             guardGasp.start();
+            contRef.GuardTalking("Ghost! WHAT IS A GHOST DOING HERE?");
```
TheBeesSneeze, line 44: Can this be a design variable?

Geist Heist/Assets/Scripts/Guards/Speechbubbles/GuardSpeechBubbleBillboard.cs
```diff
@@ -0,0 +1,50 @@
...
+    public override void OnInitialize(GameObject sourceGameObject)
+    {
+        group = GetComponent<CanvasGroup>();
+        guard = sourceGameObject.GetComponent<GuardController>();
+        guard.VoiceClipPlayed.AddListener(UpdateBubble);
+        guard.VoiceClipStopped.AddListener(stopTalking);
+        group.alpha = 0;
+    }
+
+    [SerializeField] TMPro.TMP_Text Textbox;
```
TheBeesSneeze, line 25: can this be moved to the top of the script?

Geist Heist/Assets/Scripts/Guards/Speechbubbles/Speechbubble.cs
```diff
@@ -0,0 +1,50 @@
+using UnityEngine;
+/*
+* Contributors: Brenden
+* Creation Date: 2/4/26
+* Last Modified: 2/4/26
+* 
+* Brief Description: to display a voiceline of a guard to be able to be read aswell as heard
+*/
+
+public class Speechbubble : IBillboardUI
```
TheBeesSneeze: Can we rename this script SpeechBubbleBillboard (or equivalent, just as long as "Billboard" is in the name, for easy searching)

TheBeesSneeze (changes requested):
Havent tested yet tbh bc i know youre working on stuff but i noticed some small formatty stuff

Geist Heist/Assets/Scripts/Guards/Speechbubbles/Speechbubble.cs
TheBeesSneeze: Thought about it more, can we say that it's specifically a "Guard"SpeechBubbleBillboard in the name?
Theres a lot of speech bubbles in our game, so the name is kind of vague.

TheBeesSneeze (approved):
Passes the bella test 👍

JBBateman (approved):
Good to go

PR #262, opened 2026-02-05
OOO-644

BBurtz:
When leaving a guard vision range while in a car they won't chase you to the ends of the earth anymore, this can be tested in final parlor room scene with the first car.


JBBateman (changes requested):
For some reason when the vision cone leaves a vase the guards enter vision break, probably just a variable misplaced. I don't think this is what we talked about me fixing but let me know if I'm wrong.

JBBateman (approved):
It works now

SkyBeal (changes requested):
Don't know if this is a general bug or just exclusive to this branch but I can't get caught in the car at all anymore (even when I don't escape the vision cone)

SkyBeal (approved):
can get caught in the car now!

PR #268, opened 2026-02-10
Ooo 439

Sulphide2:
The Library level grey box finished.


CalebtheMann (approved):
Approved with comments:

1. The opening space before the first interaction with guards can be shortened
2. You can just walk by the guard that's blocking the first hat (the guard cone is not long enough to cover the entire area, you can either position the guard differently or make the tables/chairs take up more space so the player has less space to work with)
3. Good job fixing the long hallways part: I DID need to use possessables!
4. The final car Moment still doesn't feel right... The room is incredibly large, the player can't see the guard timing due to the bookcases being in the way, and the hat might be impossible/require lots and lots of luck to get.
5. All hats and Tethers still need to be placed

JBBateman (approved):
Didn't break anything and I was able to complete the level

PR #269, opened 2026-02-10
OOO-450

TheLastAce:
(sorry, I couldn't change branch name)

Making Level 4 (Kitchen Wing 2) greybox
greybox is not done and there are a few guard bugs still but it is in a playable state. 
Should be no merge issues since it's in it's own scene


CalebtheMann (approved):
Approved with comments:

1. All possessables work now woot woot! (just need to add tether at the end)
2. Stationary Guard still on the table lol
3. Biggest concern after playing through is that I often found myself not needing the possessables. This is likely due to the fact that the level now feels a little too big... will have to adjust sizing at some point
4. Edges and uneven flooring/clips
5. Could you add a wall at the beginning so the player can't leave the scene lol

JBBateman (changes requested):
The level itself looks good, though I was able to complete 90% of it without possessing anything. Only requesting changes because Ollie isn't emissive and I want to take a look at it soon before getting this merged with working.

TheLastAce:
> Ollie isn't emissive and I want to take a look at it soon before getting this merged with working.

Don't worry about the emission. I turned it off for my own sanity (lowered the emission slider on the material) I'll raise it again on my end and if it still has a conflict, feel free to take from working instead.
I'm gonna keep balancing with the possessables too.

JBBateman (approved):
Emission can be fixed later.

PR #270, opened 2026-02-10
OOO-595

breyes09:
Tutorial Hallway Greybox is finished, the scene is in the AlphaGreyboxScenes folder in Scenes, and is named TutorialHallway. There is no interactable at the end of the level currently to send the player to the hub, but the level is finished outside of that.


CalebtheMann (approved):
Approved with comments:

1. This feels like a MUCH more natural tutorial than putting all of the possessables in one location
2. Temp Textures and decor looks fine and gets vibe across
3. Can feel a little blindsided by the first guard if i dont watch him walk into the room... maybe adjusting it so that the player can always see the guard, even when he walks into the room?
4. The stairs are very rough, but I believe thats a programming thing
5. Also needs interactable at the end, but you noted that

JBBateman (changes requested):
The level itself looks good but I can hardly climb the stairs. It might be that the stairs need to be less steep or the slope movement script needs to be adjusted.

JBBateman (approved):
Looks like the slopes are a programming issue

PR #271, opened 2026-02-10
Ooo 442

DrewForYou:
Wing 3 Level


CalebtheMann (approved):
Approved with comments: 

1. First off, this FEELS fun! Great work here! 
2. Smaller nitpick things: The doors don't all open up the same way, preferably if they were all pushed open that would be great (instead of some being "pulled" open)
3. Bouncing from vase to vase in the hallways while looking out for the correct door is a fun gameplay loop, but the middle area still feels the least solid to me. Perhaps expanding the size and making it into a kind of "lounge" connecting the two hallways with rooms would be something to look into. For reference, think about how a dorm room typically has a communal lounge on the center of the floor. That would provide more interesting ways to use the vending machine and add some interesting-looking props.
4. Change location Tether sends to "GlobeInside"

JBBateman (approved):
Looks good to me

PR #272, opened 2026-02-10
OOO-652

TheBeesSneeze:
Key UI.
Key UI is hidden if no keys are in the scene. Only the keys present in the scene show up in the UI.

image


JBBateman (approved):
Looks good

BBurtz (approved):
Looks good and works well!

PR #273, opened 2026-02-10
Ooo 597

jtkelly18:
Added controller support for player movement and interaction; menu navigation still needs to be implemented.


JBBateman (changes requested):
Controls quite well on my Xbox controller. The car is almost unusable with how sensitive it is to rotation with the stick, once that's altered this will be an approval.

JBBateman (approved):
Looks good

TheBeesSneeze (approved):
It works, felt good. UI needs work.
At some point, it would be cool if you could implement the Input Switch detection script i sent you, to clean the code up. But it low key works fine right now.

jtkelly18:
> It works, felt good. UI needs work. At some point, it would be cool if you could implement the Input Switch detection script i sent you, to clean the code up. But it low key works fine right now.

I can do this when I get UI working as well.

PR #274, opened 2026-02-10
Ooo 642

TheBeesSneeze:
FUNCTIONAL (it should not look pretty right now) hub level select kiosk rework.


JBBateman (approved):
Looks good

BBurtz (commented):
Looks good but are the hat slots also supposed to light up when you collect them just like the big orb for the tether?

TheBeesSneeze:
> Looks good but are the hat slots also supposed to light up when you collect them just like the big orb for the tether?

Yea

BBurtz (approved):
It currently only works for the jester hat but I assume that's because of the other 2 are currently place holder, Thanks for getting it looked at so fast!

PR #275, opened 2026-02-10
Ooo 601

JBBateman:
Intro cutscene working with a placeholder video for now. You can skip it by pressing space twice. The first space press should activate a UI element in the bottom right corner for 5 seconds, during this window pressing space a second time will skip the cutscene.


BBurtz (approved):
Looks good, Hopefully we don't use that video but the system works

jtkelly18 (changes requested):
Depending on the resolution of the game, the skip cutscene prompt appears out of view.

JBBateman:
I believe the changes requested were addressed by Toby

jtkelly18 (approved):
All concerns were addressed, looks good 👍

PR #277, opened 2026-02-11
OOO-553

smnoonan:
Making design fixes to exhibit 1, wing 1. Fixed the scale on many props, including ropes, furniture, doors, vases, etc... Fixed wall heights in multiple areas. Checked for colliders, all props *should* have them now. 

 Populated an area that gained a lot of space after fixing scale.

Changed the scale of the guards to 20. Changed Ollie's hover height to 3.


CalebtheMann (approved):
Approving with Comments:

1. None of the guards were moving, but I learned that it was cuz the nav mesh was turned off in the scene. Not pushing any changes so that I can still approve it, but during the meeting could you turn the navmesh on and rebake it?
2. The Vending Machine values need to be changed to the new values
3. This rope is new but didn't have a collider !
Image

JBBateman (approved):
Looks good

SlicedBreadStudios:
I was unable to make a build from this. Not sure if that's just my computer or this branch. 
**I'm being super nitpicky because we're at a stage of being super nitpicky so we can have something awesome. Here's a bunch of visual fixes to make:**
Screenshot 2026-02-11 091429
There's a hole in the wall here with light coming through it
Screenshot 2026-02-11 092146
We can see through the top of this half wall
Screenshot 2026-02-11 103915
This hat rack feels so small in comparison to everything else
Screenshot 2026-02-11 103924
The bookcases don't reach the ceiling here but do in most rooms
Screenshot 2026-02-11 103946
This wall change looks awkward on this corner. Add a wooden beam to the wall to make the transition smoother. 
Screenshot 2026-02-11 103955
Same with the wallpaper transition here. Add a wooden beam to that corner
Screenshot 2026-02-11 104056
Add a wooden beam here
Screenshot 2026-02-11 104139
The wall paper here is squished
Screenshot 2026-02-11 104155
The paintings are visabbly off the wall
Screenshot 2026-02-11 104243
He is tall enough to go over this bench but it feels awkward. Either he needs to be a little taller, the benches need to be a little bigger, or the benches need to be a little smaller. But this needs to make more visual sense without looking too hard. 
Screenshot 2026-02-11 104336
Gap between walls
Screenshot 2026-02-11 104413
These bookcases are tall, that's great, have the ones that aren't as tall match these ones. 
Screenshot 2026-02-11 104431
Break up this wall transition with a beam
Screenshot 2026-02-11 104546
I can go over the earlier benches but not these ones. 
Screenshot 2026-02-11 104610
The chandelier looks too into the ceiling? Might just be me on this one though
Screenshot 2026-02-11 104717
Bookcases need to be taller
Screenshot 2026-02-11 104801
I was able to go this far over this rope before hitting a collision box
Screenshot 2026-02-11 104856
Weird wall gap
Screenshot 2026-02-11 104907
Otherside of weird wall gap
Screenshot 2026-02-11 104951
I see the vision and I think this area still needs some work. It feels too busy and too empty at the same time? I'm not quite sure how to fix it. I'm inclined to say let the piano take up more space and make this a space where people would listen to someone play. Some seating areas and it's ok to have open space in a music area. Maybe this is where people danced. Maybe we change the shape of the blocked off area. The switch from mostly wood to the stone area feels off. 
Also? What's the point of the new rope? I assume to teach that you can go under it, but the player has already learned that by the time they got here so from a museum standpoint rather than a gameplay standpoint it feels weird. 

Screenshot 2026-02-11 105317
No collider on this door and the ropes feel like they go way too far past this door

Screenshot 2026-02-11 105401
Assuming this is becuase it's changed places since the lighting was done and that's why it's cast in shadow. If that's not the reason then this looks weird. 
Screenshot 2026-02-11 105415
Wall gap
Screenshot 2026-02-11 105454
This feels too small and not grand to be at the end of the hallway.

SlicedBreadStudios:
Wall beams, hat rack, and book cases fixed for the newspaper boy corner

PR #281, opened 2026-02-11
Ooo 450 test

TheLastAce:
Making another PR cause my last one destroyed my scene oops

basically merging so the level can be implemented for playtest tomorrow. Still needs balancing later


JBBateman (approved):
Everything seems to be there and the game still works.

CalebtheMann (approved):
Tether is there!

PR #286, opened 2026-02-12
Ooo 683

BBurtz:
Checked with Caleb and this is how it was wanted to be for design, It currently works in flashlight scene but it should be working in every scene, there is no variables set for any of the guards in any of the scenes or prefabs so in the other scenes there will just be stationary without moving, you can change the variables in the guard controller for the rotation variables and distance under stationary guards only.


JBBateman:
Rendering issues with the vision cone that need to be investigated

PR #287, opened 2026-02-12
OOO-699

JBBateman:
Stationary guards should now be returning to their assigned locations.


BBurtz (approved):
they will no longer be lost

SkyBeal (approved):
returned correctly!

PR #288, opened 2026-02-12
Ooo 684

BBurtz:
guards will now catch players in the vending machine, the scenes with vending machines will have to rebake after the vending machines prefabs are updated to make sure the vending machine is able to be pathed infront of


BBurtz:
according to Jacob there is issues with his rendering code so this pr will be back later

JBBateman (approved):
Looks good

SkyBeal (approved):
they got me...

PR #289, opened 2026-02-12
OOO-718

jtkelly18:


JBBateman (approved):
I could not climb guards anymore so it looks like it worked.

BBurtz (approved):
Looks good, I will miss just randomly taking to the sky but this is much better

PR #292, opened 2026-02-15
OOO-690

BBurtz:
Just updated the script to make it so that there is only stuff that is needed (I removed the option for tap because we aren't using it anymore and i removed the launch direction variable because it is no longer being used the is no functionality change in this mr.


SkyBeal (changes requested):
Got errors in the vending machine script :(

JBBateman (approved):
Looks good to me

SkyBeal (approved):
works for me now!

PR #293, opened 2026-02-15
OOO-723

SkyBeal:
made the key prompt UI not freak out when looking at an interactable/actionable object

key prompt UI should do its single animation and then no more


SkyBeal:
bug got bigger, waiting on the tobester

PR #294, opened 2026-02-15
Re-imported globe scene

JBBateman:
Scene is in AlphaGreyboxScenes right now named InsideGlobeALPHA. Admittedly the scene is not a greybox and so will likely be moved later.


CalebtheMann (approved):

CalebtheMann:
Every scene there

TheBeesSneeze (approved):
The globe didnt transition to the right hub scene but i fixed that bc its Iow key an easy fix.

PR #295, opened 2026-02-15
Ooo-450

TheLastAce:
Changes to the layout.
lots of bugs like:
- keys freeze the player
- Bouncing on guards


smnoonan:
Issues with Ollie colliding with guards, using the car sent me airborne about 90% of the time while I was testing. I think there's a wonky collider on the middle table in this section too, I kept getting caught on nothing while trying to make it past the guard.

DrewForYou (dismissed):
The issue with the key freezing Ollie appears to be still present. The stationary guard that is guarding the second hat does not reset properly. Ollie feels oddly slow and the cars feel really tiny. Check the discord thread for the bug logs of this.

DrewForYou:
Also rebake the navmesh so the guards aren't going over their patrol points

JBBateman (approved):
Looks good, I went ahead and fixed the key freezing issue, the animation errors, and the guard not returning.

Afederspiel:
Guard on the table is stuck, the guard by the second collectable seems to have its vision cone glitched. Other than that I didn't see any major bugs. Also if possible I'd recommend making the locked doors a bit more visible, I missed completely them on my first few attempts of the level.

Afederspiel (approved):

CalebtheMann (approved):

PR #296, opened 2026-02-15
Ooo 673

TheBeesSneeze:
Possession HUD rework

To Test:
* Play as EVERY possessable and make sure UI looks good.
* Make sure the button prompts exist when they're supposed to.


buimwilson:
here are my notes @TheBeesSneeze 

- The stamina wheel uses the 4-quadrant texture rather than gradiating between the 4 colors
- the 'Hold Q' prompt about charging does not appear when entered inside of a soda machine (it does for the car tho)
- the in-world bar of possessable's charge has the old texture still
- The mini view of Ollie does not include his hat
- The mini view of Ollie is not cropped to a circle shape

--- kinda related but idk
- the door in the tutorial has no interact button and pulls up a kiosk screen instead of going to the lobby
- theyres no key prompt on the lobby kiosks either
- some of the items on the lobby kiosk were giant and off screen (the rightmost one was a giant circle or something)
- lobby kiosks do not play the title card when transitioning

--- not probably you but noting it down:
- the cars in the hub are not interactable
- it looks like the anchor for guard state symbols & speech bubbles is in their torso instead of their head now

buimwilson:
More notes after 1st level:

- for some reason the ollie cam sometimes has a background, then other times doesn't?
- ollie clipping out of the cam at the top sometimes

- not sure why the hub asks me for confirmation that im leaving instead of putting me into the lobby
- have the lobby kiosks only light up if they're available. its also visible that they're in the floor a bit..

- ‼️ played through the first level, did not unlock any of the other levels.

- the proportions of the UI feel good. I have yet to see the multi-key stretch in action, since im unsure which levels actually have this in. (I tried 4 of them + the test scene).

buimwilson (changes requested):
this looks amazing for alpha!! played through tutorial, lobby, & first level. went and saw key ui by playing test scene. I have not seen the UI when it has multiple keys on it.

Works great. I have 3 asks:
- Add the 'Hold q' text onto the vending machine (from the car)
- Swap the 4-corner bar for the white one to gradient colors 
   - highest energy #0F7FFF -> #56FEEE -> #FCFFB0 -> #FFD744 lowest energy
- Swap the in-world bar texture for the new one _(also gradient pretty please, just lower transparency 👀)_

buimwilson (approved):
Changes have been completed! Thank you Toby!!! :)

smnoonan (approved):

PR #297, opened 2026-02-15
OOO-653

jtkelly18:


CalebtheMann (commented):
https://drive.google.com/file/d/1yCBK_FDuwht_11Tq7yY_-HTKlRhLC-F8/view?usp=drive_link

Here's a screen recording of how it looked for me. All in all, definitely an improvement cuz I can totally see where the vending machine is pointed, but a couple things to change:

1. Make the transition between player camera to cinemachine camera a lot faster so the player doesn't have to wait in order to use the possessable
2. The biggest problem was that I couldn't see the shot trajectory and the environment at the same time. I think this can be solved by raising the camera/zooming out, and angling the camera slightly down

Those are the two main adjustments I'd wanna see made before approving this MR!

TheBeesSneeze (commented):
Code looks fine to me, but I wanna wait until calebs requested changes have been implemented.
(Soft-approval)

CalebtheMann (approved):
Works good!!

JBBateman (approved):
Looks good to me

PR #298, opened 2026-02-16
Ooo 683

JBBateman:
Fixed an issue with the new rotation customization system that caused the vision cone to render improperly. This can be tested in any scene with a stationary guard, the rotation values should now be changeable in a section called Stationary Guards Only on the Guard Controller script component.


TheBeesSneeze (approved):
Looks fine to me

smnoonan (approved):

PR #299, opened 2026-02-17
OOO 696

TheBeesSneeze:
Fixed screen freezing between level transitions.


SkyBeal (changes requested):
screen takes a long time to disappear between scenes (you can still control ollie during this)

notably going from the globe scene -> new hub scene, the screen does not go away

JBBateman (approved):
Looks good

SkyBeal (approved):
looks good!

PR #300, opened 2026-02-17
Ooo 702 Softlock fix

DrewForYou:


JBBateman (approved):
Looks good

CalebtheMann (approved):
No softlocking, gonna approve but encountered a couple general issues:
1. Enemies could see through doors (specifically Door 5), might need to set the obstacle on a diff layer or use diff tag
2. Some enemies stayed in chase statement when they saw me enter a room and wouldn't leave chase state, but would also not run into the room and get me, so they'd just spawn camp me from outside
3. The "Ollie Cam" has the Unity background on it

CalebtheMann:
No need to take more time on this MR to change these things if they're difficult fixes. Level is good enough for Alpha

PR #301, opened 2026-02-17
OOO 714 - Greybox Updates

gbrusatti:
Updates to Initial greybox, changed things with layout and making certain area's easier and less cognitively frontloaded.


DrewForYou (approved):
Needs to link back to the correct globe scene but beyond that everything is playable.

JBBateman (approved):
Looks good

PR #305, opened 2026-02-17
OOO-694

BBurtz:
It wasn't the exit sound looping a lot it was just the possessable recharge sound playing, I removed it for the time being by the request of Brooke and Caleb


BBurtz:
Thought of something I want to test

PR #306, opened 2026-02-17
Ooo 694

BBurtz:
The issue was that the sound was being started over and over again in the coroutine I made it so that it no longer does that and plays the sound correctly, I thought about how it was started after I made my first mr


JBBateman (approved):
Looks good

SkyBeal (approved):
sounds good!

PR #307, opened 2026-02-17
OOO-695

BBurtz:
When the player leaves ghost mode the moving sound now stops no matter what


JBBateman (approved):
Looks good

SkyBeal (approved):
sounds good!

PR #309, opened 2026-02-18
Ooo 723 + 738

SkyBeal:
_Things fixed:_
- button prompts no longer go back and forth violently when looking at them
- all button prompts should now appear
- globe can now show 2 button prompts or 1 depending on game completion
- new globe prefab
- new debug tool to test the ending without having to beat any levels
- globe animation adapted to HubV2

_Things noticed but not addressed in this MR:_
- kiosks in hub are not prefabs + show their outline until you look at them. this is also linked to the giant white boxes we saw in that one build with david
- doors do not have a prefab. I did not update their button prompts to try to avoid scene conflicts. **This will have to be done by designers or someone else (I can) when all levels are finished being edited.**
- new levels are not set in the save data manager to count towards completion. This is a reminder **we must do this before we make the final alpha build.** It will take like 2 seconds I promise.

_TO TEST!_
- main changes are in HubV2, test to see if the globe allows for both key prompts + one at a time (delete save data/use the debug tool to check for without + with ending unlocked)
- look at kiosks, possessables, and other items that need button prompts to make sure they appear and are the correct button
- activate the ending and see if the animation is correct for the new hub


CalebtheMann (approved):
AWESOME JOB!! Works perfectly fine!

JBBateman (approved):
Looks good

PR #310, opened 2026-02-18
OOO-705

breyes09:
Updated the HubV2 scene. The second floor of the hub has been raised, the stairs have been reshaped to fit with the new height of the second floor, the possessable room has been shut off, all kiosks take the player to the correct level when interacted with, all doorways have been shut off with invisible walls to prevent the player from being softlocked, the new globe model has been imported and applied to the prefab, and the globe can be interacted with to allow the player to enter the actual globe inside scene. A backup of the hub scene named HubV2.5 has also been created in case there are merge conflicts to the HubV2 scene. Also made minor changes to sizes  of objects in Hubv2.5, connected TutorialHallway and GlobeInsideALPHA to hubv2.5 scene, updated first section that includes a guard of tutorial hallway to make guard more easily visible at all times, readded physics material to TutorialHallway stairs, updated main menu and pause menu to lead to Hubv2.5, and updated intro cutscene to lead to TutorialHallway


CalebtheMann (approved):
W Job!! 

However, all of Design needs to go into their levels and change where their tether sends them

JBBateman (approved):
Looks good, I'm gonna look at the rotating stuff myself

PR #311, opened 2026-02-18
OOO-725

JBBateman:
Fixed an issue where stationary guards did not have proximity aggro bubbles.


SkyBeal (approved):
touched the guards and they aggroed 👍

TheBeesSneeze (approved):
Made my quick lil fixes, looks good

PR #312, opened 2026-02-18
OOO-726

JBBateman:
Fixed an issue where the guard vision cones would distort when rendering against walls and other objects.


TheBeesSneeze (approved):
one line of code was all it took to move mountains ig

SkyBeal (approved):
don't necessarily know what i'm looking for but the vision cones look good!

PR #315, opened 2026-02-19
OOO-740

JBBateman:
Pull a stationary guard away from its post and have it enter vision break and return to its default state. It should now return to the proper side to side vision swing.


SkyBeal (approved):
they don't freak out now!

TheBeesSneeze (approved):
One line to move mountains

PR #316, opened 2026-02-19
OOO-664

TheBeesSneeze:
New button prompt sprite and font :D


SkyBeal (changes requested):
mostly looks good! I did notice that the default button text gets replaced live and looks a little weird. If it's too much to fix lmk and I'd feel alright approving this

(sorry it's so quick, the E turns to a Q in this vid. It's a bit more obvious in engine)

https://github.com/user-attachments/assets/85d137c3-46ef-4918-84f3-991abd066eea

SkyBeal (approved):
minor thing to think about later but the recharge wheel and the key prompts fight each other a bit
also globe background to E prompt is small but I imagine we'll change that in the future

looks good though!

JBBateman (approved):
Looks good

PR #320, opened 2026-02-19
OOO-450

TheLastAce:
I changed the prefabs for possessables so the timer lasts longer and charges faster. (~33%)

I just changed the values on the "Possessable Object" script for max charge and timer recharge. Hope thats right. Lmk if that actually doesn't do the right thing and I'll make the changes again correctly.


JBBateman (approved):
Looks good

CalebtheMann (approved):
Fixed some general patrolling issues and tether scene transition

PR #323, opened 2026-02-24
Ooo 685

jtkelly18:
Key and door states persist when scene reloads after reaching a checkpoint, additionally keys are registered as soon as the collection method starts (no more waiting for animations)


TheBeesSneeze (dismissed):
The key needs to get saved when the player reaches a checkpoint (sorry)

SkyBeal (approved):
appeared to work for me!

JBBateman (approved):
Looks good

PR #324, opened 2026-02-24
Remove loading cards

TheBeesSneeze:
Removed the level transition cards, and replaced them with some new animations.


SkyBeal (changes requested):
can we switch the order to animation -> load scene? the transition sometimes takes a really long time and looks a little weird on the player's end

jtkelly18 (approved):
looks and works good :)

SkyBeal (approved):
looks good! thank you for your changes!

PR #326, opened 2026-02-27
Ooo 784

TheBeesSneeze:
Sound wave shader when you use the can


JBBateman (changes requested):
A ton of errors get thrown in tutorial hallway because the shader manager slot in the GameManager is empty and I didn't see a proper sub-constant to fill it with. Once that's filled in let me know and I'll review the MR again.

SkyBeal (changes requested):
The shader automatically goes off in multiple levels, I have not seen the shader work when I use the vending machines :(

JBBateman (changes requested):
I'm still getting an error from the sound wave in FINAL Parlor Room

JBBateman (approved):
Looks good

jtkelly18 (approved):
The shader works well but might need some tweaks to make it more visible with the current lighting changes

SkyBeal (approved):
I feel like the kid from the polar express because I can finally see it now!

PR #327, opened 2026-03-01
Ooo-827

TheLastAce:
The only thing there should be is a prefab folder of the lights so designers can use them. If there are any scene conflicts take from main. I did not change anything.


JBBateman (approved):
Looks good

SkyBeal (approved):
yay prefabs!!

PR #331, opened 2026-03-01
Intentional Prop pass and slight level adjustments according to playtest feedback

Sulphide2:


JBBateman (approved):
Looks good

CalebtheMann (commented):
This level is impossible to beat now. The second-to-last moment in the long hallways needs to be adjusted: the second guard's pathing is way too short, making the timing impossible to get down

PR #332, opened 2026-03-02
OOO-759

jtkelly18:
Finished setting up controller support for UI. Should handle Main Menu, Pause Menu, Level Selection Confirm, and skipping cutscenes (select). Level Selection confirm needs some visual effect to communicate which button is being selected; otherwise, it feels pretty complete (to me). 

Using the d-pad for navigation can be very janky; sometimes inputs don't register, not sure why.


JBBateman (approved):
Looks good. There'll have to be a ticket to re-configure the settings menu to work with the controller and to make sure that the toy car is less finicky in the future.

TheBeesSneeze (approved):
looks good

PR #333, opened 2026-03-02
OOO-785

SkyBeal:
New global volume (implied darkness) has been added to the main camera! Implied darkness should now apply in every scene. We'll probably have to adjust some lights + the guard flashlights later


TheBeesSneeze (approved):
Looks good

JBBateman (changes requested):
The effect for some reason only applies to about half of the scenes in the project. Notably, TutorialHallway, Wing 3 - Bedroom, and FINAL Parlor Room do not have the effect. I didn't check every scene for the sake of time so there may be a few more.

JBBateman (approved):
Looks good now

PR #335, opened 2026-03-03
Ooo 828

JBBateman:
Deleted some stuff from the files that didn't need to be there. Moved all alpha scenes into AlphaFinalScenes to unify their location in the files. Re-organized some prefab files and created some new subfolders to organize things better.


SkyBeal (approved):
game still works! file changes look good

BBurtz (approved):
I normally go though file by file but I don't think I'm going to for this one but it should be so much easier to find whatever we need

PR #342, opened 2026-03-04
OOO-849

JBBateman:
The pause menu should no longer automatically close itself in every scene.


BBurtz (approved):
looks good

jtkelly18 (approved):
approved :0

PR #345, opened 2026-03-05
OOO-843

jtkelly18:
Made changes to player movement scripts to give ollie some friction when stopping, removed max velocity (if this needs to be added again I can do so, but it was blocking speed from increasing over max velocity and there shouldn't be other external forces acting on ollie to require a velocity cap anyway afaik).

Added a "default" option to the speed command to allow debuggers to set speed to default.


JBBateman (approved):
Looks good

smnoonan (approved):
Moving feels much better

PR #351, opened 2026-03-07
Ooo 665

JBBateman:
Player detection from the vision cone should be a lot better now. Make sure to test the guards catching player in possessables and outside of possessables.


BBurtz (approved):
Looks good seems to make it a lot more consistant

jtkelly18 (approved):
Works really well gaurds feel a lot less unfair for the player.

PR #353, opened 2026-03-08
OLLIE NEW RIG AND ANIMATIONS

Blooguy2:
In this MR you will find a scene called "AnimFuncTestTutorial." This scene is a copy of the Tutorial and has a version of Ollie that has the newest rig and animations all functioning properly. Changes can still be made, but this is the major step for getting Ollie's heavily updated model and animations into the game early on so we can fix any issues later. 

Also in this MR is an unfinished and untested version of the Guard's rig and animations; those will continue to be updated.

Ollie's animator and/or entire prefab needs to replace the main Ollie prefab that is used throughout the game, and the case _should_ be so that no animator bools need to be changed in code.


smnoonan (approved):
Looks great! Very expressive, love the idle.

Blooguy2:
UPDATE - as of 3/10, this branch has the new version of Ollie (and his hat!!) implemented into the prefab used universally for the player

JBBateman (approved):
Looks good

PR #354, opened 2026-03-08
OOO-676

JBBateman:
Exposed a variable on GuardController call StartIndex. StartIndex allows designers to tell the guard which index on a patrol path it is starting on.


BBurtz (approved):
looks good and the game is still playable

Geist Heist/Assets/Scripts/Guards/Behaviors/PatrolMovement.cs
jtkelly18, line 1: Made a change to this script, but otherwise approved

jtkelly18 (approved):
it works, doesn't break, approved.

PR #355, opened 2026-03-10
OOO-857

JBBateman:
Fixes bug where the second loading screen would never disappear.


bboster (approved):

BBurtz (approved):
you don't get stuck on the loading screen

PR #356, opened 2026-03-10
OOO-773 new wallpaper

Afederspiel:
Added a new wallpaper/fixed the tiling for it. Also I forgot to ask earlier but is the floor texture ok? I know we wanted to change it from the checkered one since it was too high resolution


JBBateman (approved):
Looks good

smnoonan (changes requested):
I do like the floor, and the tiling on the wallpaper looks much better. There seems to be an issue with the stairs on the left, there are two sets colliding with each other, so they're jittering. Could you get rid of the set of stairs underneath? I believe that's all.

smnoonan (approved):

PR #357, opened 2026-03-12
OOO-830

BBurtz:
I made it so the car wind sfx stop when you unpossess.


JBBateman (approved):
Looks good

TheBeesSneeze (approved):
One line rap god

PR #360, opened 2026-03-12
OOO 611

gbrusatti:
Totally forgot to do clutter pass and lighting on different branches so they are all on here.


JBBateman:
Github deleted a ton of work from this branch so I'm gonna have to recover this scene

PR #364, opened 2026-03-23
Ooo 804

smnoonan:


TheBeesSneeze (approved):
The level is playable with no obvious bugs.

JBBateman (dismissed):
Looks good

SlicedBreadStudios (approved):
Approved but there are things to look at. 

**This build failed**

There's no voiceline at the start and then both play at the second voice line trigger

There's an overuse of paintings. The first room has too many on the walls and the small save areas don't need them. 

All the audios play at the 3rd audio trigger. 

Ollie can walk through the rope and door in for the last hat.

The door frame in the cowboy hat room doesn't extend through the door area. 

Image
Image
Image
Image
Image
Image
Image
Image
Image

PR #366, opened 2026-03-24
Ooo 860

breyes09:
Guard balancing changes, changed move speed during chase state and amount of time it takes to leave chase and surprise states


JBBateman (approved):
Looks good

CalebtheMann (approved):
Approved for the sake of getting things out of MRs and into Working. The player is still able to run right through guard vision cones due to the updated speed, so we will still be working on finetuning some of these things

CalebtheMann:
Also, the tutorial needed its NavMesh rebaked, but that might just be a "this branch" issue

PR #368, opened 2026-03-24
OOO-872

SkyBeal:
one boolean to move mountains (+naughty attributes to make it better)

Bella declares NO confirmation menu for the globe, it will just fade to black


JBBateman (approved):
Looks good

BBurtz (approved):
Mountains were moved

PR #370, opened 2026-03-24
OOO-896

BBurtz:
Guards now stop when they get close enough


SkyBeal (approved):
not fully sure what I should be looking for but the guards reacted to the soda can sound stimulus!

JBBateman (approved):
Looks good

PR #372, opened 2026-03-25
OOO-860

breyes09:
Changed stationary guard values to feel better (increased surprised state time and time until entering vision break) and more consistent with patrol guards, and raised toy car rotation speed. Also used override on vending machine in lv 1 to set prefab values equal to lv 1 vending machine values to make them more consistent


JBBateman (changes requested):
Some of the guards in the levels are turned sideways again.

CalebtheMann:
Also, with the guards not going all the way to the source, Level 1 was unbeatable

breyes09:
Made changes based on requested change and comment

CalebtheMann (approved):
Level 1 now beatable!

CalebtheMann:
These were (mostly) guard changes, and the guard can't be touched until the animations are in. Brayden has exported the Guards as packages though on his PC, so he has all of his changes and values saved

PR #375, opened 2026-03-25
Ooo 850

jtkelly18:
This MR should make scene switching via the debug console easier while adding scroll functionality to the debug window.


JBBateman (changes requested):
Using the debug console to go to the main menu leaves the screen black, probably the fade element not working properly with the system.

JBBateman (approved):
Looks good

CalebtheMann (approved):
Idk if you were the one who did it, but the detach camera command is fire!

PR #382, opened 2026-03-26
Ooo-827

TheLastAce:
Level 1 lighting redone. Please use LevelLighting for future bakes
May need to be remerged if we decide it is too dark seen on other screens
I changed the Ollie material so he can glow again


JBBateman (approved):
Looks good

CalebtheMann (approved):
SOOOOO GOOD!

PR #390, opened 2026-03-28
Ooo 661 animations testing

BBurtz:


JBBateman (commented):
Looks good

CalebtheMann (approved):
Approved

SkyBeal (approved):
stationary guards are jank but we're working on it
one guard to the right of the lion is broken (not a prefab, replace him)

everything else was really good! might suggest an attack/catch animation though?

PR #392, opened 2026-03-29
Ooo 860

breyes09:
Guard values changed, vending machine prefab updated so values are consistent across scenes, and car rotation speed raised


CalebtheMann (approved):
Approved! Also changed the flashlight cone, so the next reviewer should see whether or not it feels fair

smnoonan (approved):

PR #393, opened 2026-03-29
OOO-898

smnoonan:
Door and plaque updates: changed the doors from green cubes to the wooden doors, moved plaques to the sides

Minor fixes, walls that were clipping, tether placement, door swing direction

Need skillet and lobster hat in engine to update tether & third collectable


CalebtheMann:
Screenshot 2026-03-29 012028
Could we add a bigger intentional prop here and tag it as "Obstacle"? This and the pillar are supposed to be things that the player can hide behind while the guard is doing his patrol path, giving the player a safety area that doesn't require possessing. Still gonna approve since it's definitely still beatable though, just something to do after this gets merged in

CalebtheMann (approved):

JBBateman (approved):
Looks good

PR #395, opened 2026-03-29
Guard Model Accessories Updates

Blooguy2:
In this PR the guard model has its flashlight, cap, and name badge, which was disabled in previous test versions. A few notes:

- Guards will need to have their lights tweaked to originate from the flashlight instead of being linked to the same origin as their red detection radius. This can be done by having the light that they currently have disabled and the light that is attached to their flashlight in the model enabled, though it might need to be tweaked. 

- The guards' accessories SHOULD be enabled in all of its animations. If it is not enabled in any particular animation or scene, it should be noted.


JBBateman (changes requested):
I see the accessories on the stationary guards but not on the patrol guards. Hopefully a quick fix.

smnoonan (changes requested):
Stationary guards are in constant running animation

Blooguy2:
> Stationary guards are in constant running animation

whoops, this was because i hadn't pulled from main in a second, seems to be fine on my end now

smnoonan (approved):

CalebtheMann:
Visually, it's pretty jank due to the spotlight being childed to the temp flashlight currently. Someone should fix it in this branch, then I'll give it my approval

JBBateman (approved):
Everything is there now.

Blooguy2:
> Everything is there now.

:D

PR #396, opened 2026-03-29
OOO-937

BBurtz:
it now uses raycasts make sure the wall is tagged with obstacle before you say it isn't working


JBBateman (approved):
Looks good

smnoonan (approved):

PR #405, opened 2026-03-29
Ooo 756

TheBeesSneeze:
Main Menu done!!
Theres more polish work to be made for it but for now it works


JBBateman (changes requested):
Some of the assets were missing from the main menu when I loaded it.

buimwilson (changes requested):
**Title Screen (initial showcase)**
- I have the same assets rendering problem as Jacob. It affects the Title and the fog PNGs.
- Can you check the looping of the squash n grows? it feels like it freezes and jumps a bit every 6 seconds.
  
**Title Screen (buttons)**
1. Can you make the time until the buttons of the title screen go back to the showcase initial screen as like... 15 seconds instead of 2?
2. If I go into the settings screen, then back into the title (buttons) page, the settings button no longer lights up with the hover state for a little while.
  
**Credits**
1. The proceed button does not light up with a hover state.
  
**Settings**
1. The Inverted X/Y checkboxes have their hover state stick until you click on something else rather than stopping once you stop hovering
2. The knobs for all the sliders do not light up on hover, only on selecting it. Could we do hover as well?
3. The longer bars for Audio have their full-states end early then tile to repeat.
Image
  
**Confirm screens**
1. Can you please have the buttons closer together towards the middle? To feel more like the buttons are connected than it being closer to the left-align of the text?

buimwilson (approved):
New changes look good to me! The edits look good and the assets are visible now.

CalebtheMann (approved):
GAS!!

smnoonan (approved):

JBBateman (approved):
Looks good

PR #406, opened 2026-03-30
OOO-939

smnoonan:
Tea pot tether and lobster hat in and working


CalebtheMann (approved):
Wait mb, updating my review

CalebtheMann (dismissed):
Nothing is broken, but I don't see the teaset model or lobster... is this hat supposed to be where the lobster is?

Image

CalebtheMann (approved):

JBBateman (approved):
Looks good. I couldn't check the hat locations because I don't know where they're supposed to be so I trust Caleb's review.

SkyBeal (commented):
commenting to say they are both in the level for me! but also the tether in the globe does not correspond to the tether in the level (I believe we know about this/are working on it, just wanna document it in case)

PR #408, opened 2026-03-30
OOO-845

TheBeesSneeze:
Fixed bug where order of keys would change


JBBateman (changes requested):
It didn't work :(

SkyBeal (approved):
all keys were working for me besides the one design bug!

JBBateman (approved):
Looks good. Assuming the mismatch in the final level is a design issue.

PR #410, opened 2026-03-31
OOO-800

SkyBeal:
globe new animations + cameras!
intended viewing is in the newfinale scene since hubv2.5 has not been stitched together with the tutorial
toggle on debugalwayspossessable on the globe for easier testing

Ending should:
- [x] be started with Q, go to first camera
- [x] globe starts swinging with the first Q press after possession
- [x] hit Q 10 times, transition to second camera
- [x] globe goes through doorway, switch to last camera
- [x] go to ending cutscene when globe goes through the final door


JBBateman (approved):
Looks good

CalebtheMann (approved):
Approved

PR #411, opened 2026-03-31
Ooo 938

jtkelly18:
Collectible system rework, added offests for wearable and display models for collectibles, added button prompts, fixed hat switching.

How to use:
1. Drag a CollectableTemplate prefab into scene, swap out default "CollectableModel" for the hat model you want to use.
2. Save the Collectable to a new prefab under prefabs/collectables/OllieHats, give the prefab a unique name.
3. Use the "Add new collectable to Registry" button to update the CollectableRegistry entry (this can be verrified by manually oppening the asset.

4. When adding to a CollectableDisplay, pick the collectable using the dropdown menu and hit preview collectable, then the CollectableDIsplay object itself can be transformed alongside with the hat, or the hat model childed to the display object can be transformed then the "Save Scene Mesh as DIsplay Offset" button can be used to update the transform saved by the registry (not sure how nessecary this button is tbh but it doesn't hurt anything :p)

5. Hat transform can be edited on Ollie by selecting the collectable from the dropdown menu on the "Wearable Collectable" script on the Wearable node on ollie. Hit the preview button, then the model itself that is childed to wearable can be transformed; DO NOT MOVE WEARABLE!!! After the model has been transformed the way you want you can hit the "Save Scene Mesh as Wearable Offset" button.


DrewForYou (approved):
Fixed issue so default hat should be correct

SkyBeal (commented):
looks good programming wise! ping me when the design portion is done and I'll come back to review :]

SkyBeal (approved):
ok ya looks good

PR #412, opened 2026-03-31
Ooo 879

BBurtz:
cans despawn
cars no longer collide with the player


SkyBeal (approved):
both worked for me!

JBBateman (approved):
Looks good

PR #414, opened 2026-03-31
Ooo 907

DrewForYou:


JBBateman (changes requested):
Hats get mixed up, seems like they may be pointing to hats from other levels. Ex. lobster and broccoli resulting in news cap and cowboy hat.

JBBateman (approved):
I CAN WEAR THE LOBSTER NOW

CalebtheMann (approved):
APPROVED!!!! (putting in the ReadMe that the Cowboy hat is the one and only bugged hat) NVM not going in ReadMe!

PR #416, opened 2026-04-07
Ooo 1000

TheBeesSneeze:
Added fog animation to main menu
Better controller support on main menu
Animation is less jittery.
And other small polish :)


SkyBeal (changes requested):
- Settings button doesn't reset to neutral after pulling up the settings menu with mouse
Image

- I think there's one automatic button click that moves the main menu? I don't click any buttons but it acts like I do (only once though (and seemingly inconsistently?))

- controller doesn't work for new game or play game 

- plugging in controller with settings already open selects the main menu buttons (play game, new game, settings, quit) instead of the settings buttons

JBBateman (changes requested):
I can't quit the game from the pause menu when using controller. It also seemed like if I deselected a button by clicking a blank space with the mouse that the controller could no longer navigate the menu.

JBBateman (changes requested):
If I click in a blank space on the pause menu it still prevents the controller from being able to navigate the pause menu anymore.

CalebtheMann:
My mouse stays on the screen permanently :/

SkyBeal (approved):
All of my reported problems are gone! This looks really good!!

JBBateman (approved):
Looks good

PR #418, opened 2026-04-07
OOO-1028

JRavioli:
Fixed dialogue in the hub (Voicelines only) so different lines play on each night up to night two


JBBateman (changes requested):
Hub needs to have line lengths adjusted.

JBBateman (approved):
Looks good

CalebtheMann (approved):

PR #419, opened 2026-04-07
OOO-962

BBurtz:
you can no longer possess through the wall anymore. mainly seen in level 5


SkyBeal (approved):
one layer mask to stop walls or something

JBBateman (approved):
Looks good

PR #420, opened 2026-04-07
Ooo 1026

breyes09:
Added car walls to levels 1 and 5 to prevent the player from clipping out of bounds, and fixed a spawn point issue in the room with the car in level 5 where the player would spawn in a wall


JBBateman (approved):
Looks good

CalebtheMann (approved):
Approved

PR #422, opened 2026-04-07
OOO-971

smnoonan:
Clutter pass 3 of Exhibit 1 Wing 2


JBBateman (changes requested):
I could not interact with the tether at the end of the level

CalebtheMann:
Also, the vending machines were in the ground and some of the bookcases didn't have colliders. We'll likely not include this MR in the FUSE build just because we've tested the other Level 2 more

JBBateman (approved):
Looks good, though I think the fog needs to be turned down.

breyes09 (changes requested):
Car walls need to be added to prevent the player from clipping out of bounds using the car

SlicedBreadStudios:
I am actively working in this branch and will make a new MR when finished

PR #423, opened 2026-04-07
Ooo 1033

DrewForYou:
Redoing the hat enum values

Test in collectibles and wearibles.unity - Jacob


JBBateman (dismissed):
Level 1 looked like it worked but level 4 did not, if this is intended then this MR is fine.

SkyBeal (commented):
don't know how in depth this PR is meant to be but some of the hats don't show up properly in the globe/display the wrong hat (the lobster hat is not currently wearable)

Image
this is the current globe scene after you get all hats in every level


this does look a lot better than the hats before though! if the goal was just to tidy up the enums to be clearer and not a total clean up of hats, ping me and I can approve this

JBBateman (approved):
Looks good

breyes09 (approved):
Looks good!

PR #424, opened 2026-04-08
OOO-1042

JRavioli:
PA voicelines now have a filter over them
New music
- Intro music
- Wing 2 music


JBBateman (approved):
Looks good, we'll have to make sure to implement wing 2's music so that it plays instead of the first level's

CalebtheMann (approved):
AWESOME!! She sounds like a real intercom...

PR #425, opened 2026-04-08
OOO-1041

breyes09:
All walls in level 1 are all now on the obstacles layer, and Player Manager prefab changed to include obstacles in Stop Exit to prevent player from clipping through walls when exiting the car


gbrusatti (approved):
Looks good!

JBBateman (approved):
Looks good

PR #427, opened 2026-04-08
OOO-960

SkyBeal:
fixed transition confirmation objects covering buttons in different resolutions! (mainly try 4K for FUSE)

the transition cards are a little ugly rn but that should be fixed in a later ticket, main reason to not do that right now is because of the hat enum PR. I did as much setup as I could for now


BBurtz (approved):
Seems to fix it

JBBateman (approved):
Looks good, handled 128x128 and Wii resolution fine.

PR #431, opened 2026-04-09
OOO661 Guard Animations Tweaks & Fixes

Blooguy2:
- Patrol and Stationary guards have had their animations tweaked and reworked for the flashlight to accurately portray their vision cone in front of them
- Flashlights across the board are significantly less bouncy
- Guard transition between animations more smoothly
- Ollie transitions between walking animation and idle animation more smoothly

Problems
- Guards _sometimes_ still get stuck in running animations once they start but I do not know if this is known already


JBBateman (commented):
I'm going to look at this in class tomorrow so I can focus primarily on it.

breyes09 (approved):
Looks good!

JBBateman (approved):
Looks good

PR #433, opened 2026-04-09
Prop Animations

Blooguy2:
Possessables will rattle every 3 seconds


JBBateman (changes requested):
Hub is missing all kiosks

JBBateman (approved):
Looks good

SkyBeal (changes requested):
these animations look very cute!

two requests if possible:
-can we make it so possessables only do their animations if you're close to them? it might be a little overwhelming to have so many things on the screen moving at once
-can we make the animations stop when you've possessed the item?

Blooguy2:
> these animations look very cute!
> 
> two requests if possible: -can we make it so possessables only do their animations if you're close to them? it might be a little overwhelming to have so many things on the screen moving at once -can we make the animations stop when you've possessed the item?

I would be happy with these changes though I am unsure of the programming requirements for these. I had thought of making a couple animations for each object where the possessables have a "breathing" look when possessed and stop rattling.

JBBateman:
It wouldn't require a ton of programming work, we would just need to carve out a task for it.

SkyBeal (approved):
we should make tasks for the changes but these look so good!

Blooguy2:
Will be changing to have all possessables perform animation every 6 seconds instead

Blooguy2:
Production looked over props with 6 second intervals of animations instead of 3 and gave the OK

PR #435, opened 2026-04-10
Clutter pass

Afederspiel:
Added interactables and baked lighting. Added temporary portraits


JBBateman (approved):
Looks good

breyes09 (dismissed):
Some walls are not on the obstacles layer, so the player is still able to clip out of bounds with the car. There is a dark spot from a lighting issue near the entrance to lv 3. Outside of these, everything looks good

smnoonan (changes requested):
- Kiosks need button prompts!
- Kiosks on first floor have blue screen, while second floor have black?
- New wallpaper looks good! Want to get new flooring in eventually, just to keep in mind

breyes09:
Stairs and invisible wall in front of desk need to be put on obstacles layer to prevent clipping out of bounds

smnoonan (approved):

PR #436, opened 2026-04-13
OOO-674

JBBateman:
Scene load sequence. Should stop all sound, unload a scene, collect garbage, and load a scene.

Added OnLoadStarted and OnLoadCompleted events that are invoked during the loading sequence.


TheBeesSneeze (approved):
There was some problems with the main menu, but they were easy fixes so i just did them :P

SkyBeal (approved):
looks good to me! I think the opening cutscene is fixed in a different branch (that's the only negative thing I saw)

PR #439, opened 2026-04-14
Ooo 1081

JBBateman:
Reverts Josh's slope movement fixes to prevent a movement bug that caused Ollie to jitter back and forth in certain scenes and clip out of the map.

The scene this happened consistently in was the globe scene.


SkyBeal (approved):
I don't know which scenes were the problem, but I was not jittering or clipping in the hub or the tutorial hallway!

JBBateman:
Updated description to include the scene it happened most consistently in.

BBurtz (approved):
Can't find a way to clip out so looks good to me

PR #440, opened 2026-04-15
Ooo-966

TheLastAce:
Clutter pass. It's def not done so Sophia will help me finalize it. Just make sure no game breaking issues with the props used.
Also, hat registry.


JBBateman:
I'm gonna close this until it's finalized so that we don't accidentally review and merge it early.

PR #445, opened 2026-04-16
AddedKiosk&PASystem (Not Removing Prefabs)

BlazingBlitzi:


JBBateman (approved):
Looks good

CalebtheMann (changes requested):
For the kiosk, can you separate the screen material from the actual model material? Similar to how you had the first kiosk, that way we can just slap a material onto the "screen" in order to turn the kiosk off, instead of meticulously creating a UI pane and trying to perfectly place it on top of the screen. Other than that, no bugs and the model looks good!

CalebtheMann (approved):

PR #446, opened 2026-04-16
OOO-1035 & OOO-1038

SkyBeal:
ollie movement particles, possessable particles, and tether vfx implemented!


JBBateman (approved):
Looks good

CalebtheMann (approved):
Approving for the sake of getting it in, but AS A NOTE: The hats don't have particle effects, and we should recenter the effect on the candle holder

PR #449, opened 2026-04-16
Ooo 1111

DrewForYou:
Made sure all hats look right in the globe


DrewForYou:
Make sure to go through ALL the levels the player will go through and collect the hats to make sure something doesn't overwrite the model of one. Do note the RED lobster in level 3 doesn't have it's enum attached yet so that won't properly appear atm.

CalebtheMann (dismissed):
Alt Jester Hat, Alt Teacup, Alt Lobster and Regular Lobster, and Wizard Cap all did not appear correctly in the globe.

DrewForYou:
Lobster seems to be the only hat still not properly connected, will fix once all outstanding branches for it have been closed

DrewForYou:
Let me check to make sure the wizard hat is no longer clipping through Ollie

DrewForYou:
Yes the wizard hat is fixed

SlicedBreadStudios (approved):
Approved!!

JBBateman (changes requested):
I'm having the same issues that Caleb mentioned earlier.

JBBateman (approved):
Works and remaining things will be fixed later.

PR #450, opened 2026-04-16
OOO-1130

JBBateman:
Guard flashlight now changes color between states. The only differing color I actually set up ahead of design is surprised and chase state turning the light bright red. You can change a var called Flashlight Color on each behavior to test if it properly changes with behavior changes.


CalebtheMann (approved):
Works great! Where's the behavior at though that Design can mess with? I lowkey couldn't find it on the Patrol Guard Controller Script...

JBBateman:
It's on each individual behavior

SkyBeal (approved):
this looks sick!! all working on my end

PR #451, opened 2026-04-18
OOO-963

JBBateman:
This should fix the lighting being really bright when you load into a scene. Let me know if it doesn't work on anyone else's end.


SkyBeal (approved):
it didn't appear bright on my end!

smnoonan (approved):

PR #452, opened 2026-04-19
OOO-1051

JBBateman:
Ollie can now be in moving state after leaving a possessable while the movement keys are held.

Also turned the anim bool string variables to constant so they can't be changed accidentally.


smnoonan (approved):

TheBeesSneeze (approved):
yea lol looks good

PR #457, opened 2026-04-19
Ooo 967

SlicedBreadStudios:
Decorations pass on Level 3
- all vases should be facing out into the room
- everything should have colliders
- All stanchions should have colliders to the ceiling


smnoonan (approved):

CalebtheMann (dismissed):
Works fine but there are SOOOO many changes... including light map things. Are all of these changes accurate? If they are, then we can remove the "changes requested" and approve this

SlicedBreadStudios:
Lobster hat is good. I moved every lighitng fixture up so that should be why the light maps are weird.

JBBateman (approved):
Looks good

PR #458, opened 2026-04-20
Ooo 1138

SlicedBreadStudios:
Set Dressing for Level 5

Known things that will need to be fixed:

- Vases will let the camera go through the wall in the hallway 
- Tether and particles are not lined up in their spot
- Doors turn the wrong way


smnoonan (approved):

CalebtheMann (dismissed):
Works fine!

JBBateman (approved):
Looks good

PR #460, opened 2026-04-21
OOO-1078

BBurtz:
guards now no longer keep running into nothing


CalebtheMann (approved):
Was working for me!

SkyBeal (approved):
they're fixed!

PR #462, opened 2026-04-21
OOO-1117 End Scene Animations

Blooguy2:
This MR contains guard and door animations for the globe rolling cutscene. All of its content can be found in the scene named ENDINGSCENEDUPE, which is a copy of HubV2.5 and is identical in all but a couple ways:

- The main doors of the Hub have been placed in an empty parent named "endingcontainer." This object contains the animations for the doors bursting open when the cutscene is meant to start.

- A object named "guardscontainer" exists with an animator that moves the guards forward. They are eternally waiting behind the main doors of the hub. 

- 6 guards named "endsceneguard 1-5" are in bowling pin formation. Each of these have their actual body animations attached to them. 

All three of these objects have a starting animation that can be triggered by a bool named ENDSCENESTARTED. When this is toggled true, their animations start. The guards will run in, and they are intended to loop their "Looping" animations until the globe hits them, which then should trigger their "Death" animations. 

For an idea of how the ending movements look you can simply press play in "ENDINGSCENEDUPE and watch the animation play out. 

Let me know if any more changes or help is needed with getting these animations to function properly!


Blooguy2:
oops i did the wrong thing

Blooguy2:
Just lost all of the work I did today in this branch. Will need to fix

Blooguy2:
OK IT'S ALL GOOD. It wasn't even broken in the first place. Amendment to my original comment; you cannot properly preview the animations by hitting play. I will post a video later of what the preview SHOULD look like, but having it set where you can watch it doesn't line up with what's needed for programming to plug everything in.

JBBateman (changes requested):
Lighting for level 2 is deleted in this branch, I can fix it but I'm gonna have to wait until tomorrow.

JBBateman (approved):
Looks good

SkyBeal (approved):
the preview on discord looked good! i will definitely be pinging you to see how I can implement these though lol

PR #463, opened 2026-04-21
Ooo 1062

TheBeesSneeze:
Implemented new art in the level kiosks!!
To test: interact with every kiosk, play the game in multiple resolutions, use the kiosk to transition scenes, test with different hat collected states (you can use the new command "c hats random" to have randomly unlocked hats!!) also just Imk if anything Iooks bad.

image


SkyBeal (changes requested):
this looks sooooo good. I have a couple of nitpicks but if you decide some of these are too tedious/difficult, ping me and I can come back to approve. 

things to change:
- free aspect and 16:10 has some weird looking stuff (particularly text). not unbearable, just could be tweaked a _bit_ more

Image

Image

- tethers save when you enter a level and are marked as completed even if you haven't completed the level **(I think I have this fixed on a PR that's about to drop)**

- some punctuation stuff isn't consistent for the blurbs (2nd missing an ', 3rd has no ending punctuation, etc.)

- hat on left side of level 5 kiosk is getting clipped off

Image

things to add? :
- I think it could be cool to have a different color of light on the collectables bc they seem very cold rn

SlicedBreadStudios:
Merge OOO-1111 before this one

Sulphide2:
The Kiosks disappear after finishing a level, so that needs to be fixed for this to work. The transition changes only work using the command because of this since you can't see the transition screen more than once.

TheBeesSneeze:
> The Kiosks disappear after finishing a level, so that needs to be fixed for this to work. The transition changes only work using the command because of this since you can't see the transition screen more than once.

Can you reword this? I dont think I understand what you mean.

SkyBeal (approved):
absolutely stellar. everything is working on my end! free aspect the enemy of all

CalebtheMann (changes requested):
After some testing, I encountered a similar issue to Alex, but I don't think it's your system. Adam made it so that certain objects are set to turn on and off in the hub when certain things happen like it's intended, but instead of turning the kiosk screens on and off, the whole object is turned on and off (and a little inconsistently, but he's still working in the hub). 

I requested changes though cuz of these . . . The possessable UI is now just a square for all of them :/

Image
Image
Image

CalebtheMann (approved):
WOOT WOOT!

PR #464, opened 2026-04-22
OOO-966

smnoonan:
Clutter pass 5 of lv 4
- made changes to the first guard interaction so that the guard would not turn and see the player at spawn


CalebtheMann (approved):
One of the vases' models and prefab locations are separated, but I'll fix that when I slap in the dialogue tonight. Other than that, looks good!

JBBateman (approved):
Looks good

PR #468, opened 2026-04-23
OOO-1137

JBBateman:
Guards should no longer get stunned by the car or cans.


CalebtheMann (approved):
No more guard stun

breyes09 (approved):
Stunning removed, looks good

PR #469, opened 2026-04-23
OOO-1083

JRavioli:
Music switches for everything but tutorial
Tutorial music is missing until Brennan finishes it tonight


CalebtheMann (approved):
AWESOME !!

JBBateman (changes requested):
Wing1 and Wing3 events are missing in the FMODEvents scripts on my end, and this breaks the entire game. Might be since I pulled working into it, but this needs to be checked out before I can approve.

JBBateman (approved):
Looks good

PR #472, opened 2026-04-23
Ooo -152

TheLastAce:
I'm so sorry. Level 3 lighting is trying to kill us and we must abandon it. Temporary lighting is added so we can see. Please still merge this in cause idk if final lighting is possible


CalebtheMann (approved):
So sad T_T

JBBateman (approved):
Looks good

PR #473, opened 2026-04-23
Ooo 1154

DrewForYou:
Check to make sure you can't climb up on the suitcases and the leaning picture on the ground. Ollie skull should be in the globe


CalebtheMann (changes requested):
When swapping hats in the globe, the Top Hat did not appear on top of the skull. I think it disappeared cuz I couldn't find it

DrewForYou:
I know what happened, I forgot to save the changes I did to the collectable registry which is why the hat didn't change properly, I can fix that here soon

JBBateman (approved):
Looks good

CalebtheMann (approved):
W!

PR #476, opened 2026-04-26
Fix

JBBateman:
This should repair the missing work issues with working.

Note that some lighting may be missing since there were 3000 conflicted lighting files that I deleted.


SkyBeal (changes requested):
problems/things missing I noticed:
- no music in tutorial
- lights in hub are super bright
- no camera transitions in the hub (this also prevents the ending from happening)
- lighting in level 2 is super bright
- car moving noise doesn't stop if you leave the car while it's playing
- no key in level 3
- no key prompts on the globe
- wrong audio plays on the ending cutscene

CalebtheMann:
1. The camera transitions weren't placed in the hub because there's an open MR with hub changes
2. The key in Level 3 was turned off in a lighting pass, but can easily be turned back on
3. Were the lights super bright I'm those levels consistently? I find that sometimes restarting Unity fixes my lighting issues tbh

The rest are valid !

SkyBeal (commented):
key + camera are easy to add back fs

I reloaded a couple times and my lighting was still super bright :/

CalebtheMann (approved):
Let's get this in!

SkyBeal (approved):
what caleb said

PR #479, opened 2026-04-28
Ooo 982

BBurtz:
Achievements
if you load them in engine it will connect to steam so you can test it in engine you don't need to have a build for it.


JBBateman (approved):
Looks good

Geist Heist/Assets/Scripts/Player/Possession/TetherPossessable.cs
```diff
@@ -23,6 +23,9 @@ public class TetherPossessable : IInputHandler
 
     private Coroutine victoryAnimation;
 
+    [SerializeField] bool hasAchievement;
+    [SerializeField] AchievementManager.eAchievements WhatAcheivement;
```
TheBeesSneeze, line 27: Not gonna ding you for this since it doesnt really matter, but you should look into NaughtyAttributes ShowIf() tag 
https://naughtyattributes.com/attributes/meta_attributes/show_hide_if.html

Geist Heist/Assets/Scripts/Steam achievements/AchievementManager.cs
```diff
@@ -0,0 +1,102 @@
+using UnityEngine;
+using static Unity.Collections.AllocatorManager;
+
+public class AchievementManager : Singleton<AchievementManager>
+{
+    public static AchievementManager instance;
```
TheBeesSneeze: AchievementManager already extends Singleton, so it now has two instance variables. Please delete this one.

(Also, I see that youre making this script DontDestroyOnLoad, Use DontDestroyOnLoadSingleton<AchievementManager> instead of Singleton<AchievementManager>)

TheBeesSneeze (changes requested):
Can you spawn and initialize the steam achivement manager in GameManager to follow our manager pattern?

BBurtz:
> Can you spawn and initialize the steam achivement manager in GameManager to follow our manager pattern?

I need it spawned in off rip and game manager doesn't even exist on the main menu I tried this initially and then it caused issues

TheBeesSneeze (approved):

PR #486, opened 2026-04-28
Ooo 1064

jtkelly18:
Problem : 
When near a kiosk, mashing E fast enough can open multiple menus to enter the level. This also allows the pause menu to be opened over the other menus.

Additionally, mashing E on the globe can trigger the fade-to-black animation multiple times.

Same with getting caught by multiple guards.

Solution:
Set up coroutines and boolean flags to ensure that multiple “fade to blacks” or confirmation menus can’t be opened at once.


Geist Heist/Assets/Scripts/UI/Scene Transitions/FadeToBlack.cs
```diff
@@ -33,5 +34,6 @@ private IEnumerator FadeAnimation(UnityAction action, float delayAfterAction)
         yield return StaticUtilities.FadeToHidden(fadeCanvasGroup, secondsForFade);
         Destroy(gameObject);
         fadeCoroutine = null;
+        GameManager.Instance.fadingToBlack = false;
```
TheBeesSneeze: Could you set this to true in the initialize function?

TheBeesSneeze: RIght now it is only getting set to true in one specific instance

TheBeesSneeze (changes requested):
Pressing "Return to lobby" from a level has a null reference

JBBateman (changes requested):
Using the new game button with no save data throws an error and doesn't start the game at all.

JBBateman (changes requested):
I'm now getting a ton of compiler errors based around the yield return and StartCoroutine functionality for the fade to black.

JBBateman (changes requested):
It's almost done. I had to put the lines that set the bool in GameManager in an if statement since the Main Menu and Cutscene scenes don't have a GameManager. After skipping the intro cutscene the black transition did not go away for me anymore, hopefully not another silly issue that I could have fixed in two seconds if I saw it.

JBBateman:
I got a ton of compiler errors and I think this MR might be trying to overwrite people's work now.

PR #487, opened 2026-04-28
Ooo 1148

TheBeesSneeze:
Fixed bug where settings would get reset after you got caught by a guard (and the level restarts).


SkyBeal (changes requested):
settings changes only apply when you open the menu
the settings themself save though!

SkyBeal (approved):
worked for me!

smnoonan (approved):

PR #489, opened 2026-04-29
OOO-1161

gbrusatti:
Rebaked the lighting in Wing 2 Level 2 and added lights to 2 rooms that needed lighting


breyes09 (approved):
Lighting seems a bit bright in some places, but looks good otherwise

JBBateman (approved):
Looks good

PR #490, opened 2026-04-29
OOO-1160

DrewForYou:
Long list of things in ticket as well as swapping 3 and 4 tethers in the globe and in level 3, fixing hat sizes in level 2, and disabling autoregistry to stop it from messing up the hats


JBBateman (approved):
Looks good

CalebtheMann (approved):
Works great !!

PR #495, opened 2026-04-30
Ooo 1159

BBurtz:
Changed the cutscene videos and then changed how the ending cutscene works
ending cutscene now gets to the end and loops until you skip


JBBateman (approved):
Looks good

breyes09 (approved):
Looks good

PR #496, opened 2026-04-30
OOO-1198

JBBateman:
Fixed some mesh errors that showed in the console for level 3.


breyes09 (approved):
Looks good

CalebtheMann (changes requested):
Image

I'm getting this... Is it worth being concerned about?

CalebtheMann (approved):
All good

PR #499, opened 2026-04-30
OOO-1196

BBurtz:
Made it so that you can interact with flavor text multiple times and added the flavor text achievement


Geist Heist/Assets/Scripts/Save Data/SaveDataManager.cs
```diff
@@ -323,8 +323,17 @@ public int getCansThrown()
...
         return currentSaveDta.CansThrown;
     }
+    public void flavorTextRead()
+    {
+        currentSaveDta.FlavorTextRead++;
+    }
+
+    public int getFlavorTextRead()
+    {
+        return currentSaveDta.FlavorTextRead;
```
TheBeesSneeze: can you just do currentSaveData.FlavorTextsRead.count() ?

Geist Heist/Assets/Scripts/Steam achievements/AchievementManager.cs
```diff
@@ -116,5 +116,13 @@ public void TryUnlockAllAchievements()
 
         // there are more achievements missing, but i think this is the important one.
     }
+
+    public void checkFlavorText()
```
TheBeesSneeze, line 121: can you call this function in TryUnlockAllAchievements() ?

Geist Heist/Assets/Scripts/Save Data/SaveDataManager.cs
TheBeesSneeze: this is safer because im pretty sure the pIayer can read the same fIavor text muItipIe times and get the achievement.

TheBeesSneeze (changes requested):
There is nothing stopping you from mashing interract with something and having a million overlapping voice lines.
Can we make it so you can only interact with it once per scene? so like, you interact with it it once, and then you cant redo it until you reload the scene.

JBBateman (approved):
Looks good

breyes09 (approved):
Couldn't test the achievement due to not being able to interact with an interactable in the hub (this is a design bug), looks good otherwise

TheBeesSneeze (approved):
Yay! Thank you brenden!

PR #502, opened 2026-04-30
Ooo 1187

JBBateman:
Both Ollie and the Guards play an animation the player gets caught.


CalebtheMann (approved):
Approved, but is it possible to delay the catch so that players can see the full animation?

CalebtheMann (approved):

BBurtz (approved):
Looks good

PR #503, opened 2026-04-30
Ooo 1040

TheBeesSneeze:
Pause menu polish for controllers. 
When testing:
* please swap between controller and keyboard a lot
* please try to use the controller to switch between different combinations of menus and buttons.

let me know if anything feels unintuitive!! id be happy to change it.


JBBateman (changes requested):
It is possible to move the selection off of the buttons by moving sideways with the dpad in the pause menu. That's the only thing that seemed broken though.

jtkelly18 (approved):
Looks good, only thing I think would be nice is the ability to jump from the options on the right hand side back to the menu sections on the left by pressing the right face button.

JBBateman (approved):
Looks good

